PR #1, opened 2015-07-18
Fixed issue w/ already registered cells.

zachwhelchel:
Fixed issue where registering cell inside storyboard caused cell to not load properly by checking if cell is already registered to the storyboard.

### DIFF
--- a/FirebaseUI/Implementation/FirebaseTableViewDataSource.m
+++ b/FirebaseUI/Implementation/FirebaseTableViewDataSource.m
@@ -42,7 +42,9 @@
         
         self.populateCell = ^(id cell, id snap) {};
         
-        [self.tableView registerClass:self.layoutClass forCellReuseIdentifier:self.reuseIdentifier];
+        if ([self.tableView dequeueReusableCellWithIdentifier:identifier] == nil) {
+            [self.tableView registerClass:self.layoutClass forCellReuseIdentifier:self.reuseIdentifier];
+        }
     }
     return self;
 }


zachwhelchel:
Calling registerClass:forCellReuseIdentifier: on a "prototype cell" that has been setup in a storyboard can cause it to not load properly. As seen here: (http://stackoverflow.com/questions/19304528/ios-7-dequeuereusablecellwithidentifierforindexpath-method-does-it-need-regis).

Here is an example I ran into...

Custom cell in storyboard prototype:
screen shot 2015-07-18 at 5 10 58 pm

And how it loads in the tableview without the fix implemented:
screen shot 2015-07-18 at 5 11 23 pm

And how it runs with the fix:
screen shot 2015-07-18 at 5 15 41 pm

The proposed fix simply checks to see if the tableview is able to call up a cell with the given reuse identifier. If it can then it need not "registerClass" as it already is registered.

puf:
Wow. You weren't joking when you said you'd get started with the library
within a few days. Great to see Zach!

zachwhelchel:
@puf glad to help! I'll keep you all posted if I find any more potential improvements!

mcdonamp:
Yep, I've fixed it by just removing it (and the cell class all together), which [cleans up the constructors significantly](https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseUI/API/FirebaseTableViewDataSource.h#L38-L39). It means that you now have to call registerCell: or registerNib as you normally would (unless you're using storyboards, where you don't have to call it at all), but I'm open to different functionality if we think that hiding that and passing classes or nibs along would be a good thing to add.

mcdonamp:
Ok, I ceded to that this is a good idea :)

I think overall it results in an easier API if we hide this from folks, plus we can explicitly check and avoid the headache this cause both of us. Closing the issue since it's fixed here: https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseUI/Implementation/FirebaseTableViewDataSource.m#L68-L70 and here: https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseUI/Implementation/FirebaseTableViewDataSource.m#L85-L88

PR #13, opened 2015-09-30
Sectioning

pmodernme:
@mcdonamp 

A separate pull request for Sectioning. 
This version is dependent on the `NSSortDescriptors` and `NSPredicate` version.

`FirebaseArray` now has a `sectionKeyPath` property. This is a key path for use with an `FDataSnapshot`.
eg:

```
dataSource.array.sectionKeyPath = @"value.simpleDate"; 
dataSource.array.sectionsOrderedAscending = YES;
//Sections will be produced based on the simpleDate value
```

You can access the values the sections are based on by the `sectionValues` property of the array.

This version uses `NSIndexPaths` instead of `NSUIntegers` in the dataSource methods.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,0 +1,9 @@
+# This is the official list of FirebaseUI-iOS authors for copyright purposes.
+# This file is distinct from the CONTRIBUTORS files.
+# See the latter for an explanation.
+
+# Names should be added to this file as:
+# Name or Organization <email address>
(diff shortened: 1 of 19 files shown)


russellquinn:
Hi, is there likely going to be support for UITableView sections anytime soon? Thanks.

jazpearson:
I'd certainly like to see support for sections too.

PR #28, opened 2015-12-05
Raise cancelations

chrisellsworth:
The intent of 588d3ee is for consumers to be able to detect when the backing query was canceled so that errors can be handled accordingly.

e6edbb4 is not related to this change but was necessary in order to be able to build the project.


mcdonamp:
@chrisellsworth looks good to me, once you've signed the CLA I'll merge it in. Thanks!

chrisellsworth:
@mcdonamp -- thanks Mike, I've signed it!

mcdonamp:
Ok, this will go out in 0.3.2 :)

PR #29, opened 2015-12-06
RESTful fallback for large datasets

chrisellsworth:
Optionally (via `FirebaseUIOptions` parameter) fall back to RESTful fetch (with `shallow=true`) when the backing dataset is too large for the websockets API. Will auth if necessary. Can probably do some fancier things with different query parameters for ordering/filtering in the future.

There are quite a lot of constructors for the data sources now.

Is there a better way to populate an `FDataSnapshot` than subclassing it and overriding methods?

### DIFF
--- a/FirebaseUI.xcodeproj/project.pbxproj
+++ b/FirebaseUI.xcodeproj/project.pbxproj
@@ -22,6 +22,7 @@
 /* End PBXAggregateTarget section */
 
 /* Begin PBXBuildFile section */
+		83D0E7EF1C13A62A00652C4A /* FRestDataSnapshot.m in Sources */ = {isa = PBXBuildFile; fileRef = 83D0E7ED1C13A62A00652C4A /* FRestDataSnapshot.m */; };
 		BB372B551B8BE4890046905E /* FirebaseTwitterAuthProvider.m in Sources */ = {isa = PBXBuildFile; fileRef = BB372B541B8BE4890046905E /* FirebaseTwitterAuthProvider.m */; };
 		BB372B601B8D1B580046905E /* FirebaseTwitterAuthProvider.h in Headers */ = {isa = PBXBuildFile; fileRef = BB372B531B8BE4510046905E /* FirebaseTwitterAuthProvider.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		BB4877901BA8A0C000FD3D4D /* GoogleSignIn.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = BB48778F1BA8A0C000FD3D4D /* GoogleSignIn.framework */; };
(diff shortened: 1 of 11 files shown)


mcdonamp:
A few thoughts:
  1. `FirebaseUIOptions` Yes, though I'm considering a rewrite of the API when we add sectioning to incorporate a constructor to create the datasource with options, delegate, etc; then adding sections (and thus the references and their options) differently.
  2. Yeah, they are becoming too unmaintainable, hence the rewrite above. I thought `__kindof` support would help out (since I can put `__kindof FQuery *` for everything, but swift seems to not be a huge fan of `__kindof`.
  3. When you say "populate an `FDataSnapshot` what exactly do you mean? Populate a view from a snapshot, or get a model object back?

The code for 1 could look something like:
```objective-c
Firebase *ref = [[Firebase alloc] initWithUrl:@"https://foo.firebaseio.com"];
FirebaseTableViewDataSource *tableViewDataSource = [[FirebaseTableViewDataSource alloc] initWithOptions:@{@"enableRest": YES} andDelegate:self];
[tableViewDataSource addSectionNamed:@"MySection" withQuery:ref orderBy:NSOrderedAscending]; // orderBy is optional, as would be `filterWithPredicate:(NSPredicate *)pred` or similar
```

chrisellsworth:
@mcdonamp 

1/2: Fewer ~~constructors~~ initializers (sorry my Java is showing) seems like a step in the right direction. I almost fear that the new methods for adding sections will blow up in the same way though with all of the optionals and two branches to support both `FQuery` and `Firebase`. What about breaking into additional classes and then imposing on the consumer to set optionals via properties? For example:

```objective-c
FirebaseTableViewDataSource *tableViewDataSource = [[FirebaseTableViewDataSource alloc] init];
tableViewDataSource.delegate = self;
tableViewDataSource.options = @{@"enableRest": YES};

FirebaseDataSourceSection *mySection = [[FirebaseDataSourceSection alloc] initWithQuery:ref];
mySection.name = @"MySection";
mySection.orderBy = NSOrderedAscending;

[tableViewDataSource addSection:mySection];
// or tableViewDataSource.sections = @[mySection];
```
Is there any concern with breaking the API for existing consumers?

3: Apologies for being vague. What I'd like to do when `FirebaseArray` falls back to RESTful fetch is still store and return `FDataSnapshots` even if the result didn't come out of the realtime wiring, that way the fallback is transparent to the consumer. [For example.](https://github.com/chrisellsworth/FirebaseUI-iOS/blob/restful-fallback/FirebaseUI/Core/Implementation/FirebaseArray.m#L184) However, I couldn't find a way to populate an `FDataSnapshot`'s `ref` and `value` manually -- whatever the Firebase events are doing internally to create them doesn't appear to be exposed. So I subclassed it and overrode `ref` and `value`.

mcdonamp:
1/2: Yes, I think this is likely the better solution. As for breaking people--we're in 0.X, so I'm comfortable breaking people until we hit a 1.0. I think there's always the option of staying on 0.2 or 0.3 if the features in 0.4 aren't desired.
3: Reflection aka why I love Obj-C ;) (because we purposely don't expose this functionality to prevent people from mutating snapshots and storing them/trying to pass them back in). Maybe creating a `FMutableDataSnapshot` that is a container around an `FDataSnapshot` might be a nice wrapper though? That said, I'm working on strongly typed object marshaling (ala `modelClass`) in a more safe way, like Java's `getValue(Model.class)`.

chrisellsworth:
3: I generally favor composition over inheritance (a la `FMutableDataSnapshot` wrapper), though in this case it would break anyone specifically looking for an `FDataSnapshot` to populate their `*viewCell`s. Once `modelClass` is implemented though, then it does seem like this would no longer be a problem for those that use that feature (and possibly even those who don't, depending on how the API is set up). Will `modelClass` (I guess that's what I'm calling the feature now) still be able to provide the backing ref to interested parties for further inspection/drilldown? Perhaps I should hold off on development of RESTful fallback until then.

mcdonamp:
Yeah, so on second thought... `FDataSnapshot` is a little crazy and won't be easy to populate, so we might want to go for the `Mutable` version ;)

Model class is still very conceptual, and involves a lot of hacky runtime hacks (recursive introspection of types essentially), which I have no idea if it will be performant enough to do in the client. Currently it's just doing `setValuesForKeysWithDictionary` which restricts us to objects containing primitives, hence the lack of documentation. And no, it doesn't provide a backing ref, though that's still retained by the actual snapshot (which is stored), so I could see the need for having both the refs as well as the objects stored (essentially a "strong" snapshot, which actually contains a model object of some sort).

I would play around with a few features and see if anything feels good, I'm also happy to share some FDataSnapshot -> class code, if you'd like to learn the magic of Obj-C runtime hackery ;)

PR #36, opened 2016-02-07
Fixed unnecessary "cell added animation" with FirebaseTableViewDataSource.

noughts:
https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseUI/Core/Implementation/FirebaseTableViewDataSource.m#L254

If there are 1000 data, FirebaseTableViewDataSource calls beginUpdates, insertRowsAtIndexPaths and endUpdates 1000 times on initialize. That is a waste.

So I enhanced FirebaseArray initializing with observeSingleEventOfType.

Sorry for my poor English.
Thanks!

### DIFF
--- a/FirebaseUI/Core/API/FirebaseArrayDelegate.h
+++ b/FirebaseUI/Core/API/FirebaseArrayDelegate.h
@@ -40,6 +40,11 @@
 @optional
 
 /**
+ * Delegate method which is called whenever array initialized with query.
+ */
+- (void)childrenInitialized;
+
+/**
  * Delegate method which is called whenever an object is added to a
  * FirebaseArray. On a
  * FirebaseArray synchronized to a Firebase reference, this corresponds to an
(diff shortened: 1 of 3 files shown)


noughts:
I signed it!

noughts:
@davideast @mcdonamp

mcdonamp:
@noughts Thanks for taking some time to do this! Give us some time to think on this--we've had a number of comments on animation behavior (like having initial load animations, as well as having different animations for insert/remove/change) that are making us re-think how we do animations. In reality, these are developer and application specific actions, so we shouldn't be dictating how it works, but instead providing a simple framework to pass that through to the end developer. Having blocks that specify animation style for each action, or a flag that toggles this behavior would be more of what we're looking for.

andr-ggn:
So, what you are decided. I think it is good to have initialisation values.

PR #70, opened 2016-06-05
Update README.md to use signInProviders instead providers

mcz:


mcz:
I signed it!

mcdonamp:
This name change is actually what I'll refer to as "optimistically correct" in that it'll be right when we upload a 0.4.1, which is coming down the pipeline soon.

I'll close this because it will be fixed soon, but know that we appreciate contribution (even [and arguably especially] at this level), so thanks for doing this :)

Issue #67, opened 2016-06-02
Unable to build Example App


grantkemp:
I tried to build example app - but I got the following error: 
#import "FirebaseUI/FirebaseTableViewDataSource.h" not found. 
Has this been deprecated?

albertodebortoli:
Same here.

grantkemp:
@albertodebortoli  Yay - its not just me... Tell Marco hi from me. Nice to see the work you guys are doing with JE App

mcdonamp:
It'll be `FirebaseDatabaseUI/...`. We're working on updating the app with the new Authentication, so we'll clean this up along the way. Sorry about the brokenness :/

grantkemp:
@mcdonamp if you can find a way to make the demo app a single page app( or at most 2 vcs)  - it would make it really simple to understand. If you can't demo in a single page app- its not worth demoing...

mcdonamp:
I think it is a single VC, right? Is the issue that there are things spread across multiple files (there's a custom datasource, custom cells, etc.)?

dakeshi:
@mcdonamp, In addition to filename issue,  FirebaseUIChat example has more issue about podspec.
When I install the FirebaseUIChat with the following instructions in README, FirebaseUI pod is not existed in the workspace. 

> cd FirebaseUI-iOS/examples/FirebaseUIChat
pod install
open FirebaseUIChat.xcworkspace

If I edit Podfile with `pod 'FirebaseUI'` instead of `pod 'FirebaseUI', :path => "../../FirebaseUI.podspec"`,
I can find out FirebaseUI in the workspace.

I think that there are some problem in FirebaseUI.podspec, especially the below line:

>   s.source       = { :http => 'https://dl.google.com/firebase/firebaseui/ios/0_4_0/FirebaseUIFrameworks.zip' }

So, I edit podspec with this:
>   s.source       = { :git => 'https://github.com/firebase/FirebaseUI-iOS.git', :tag => s.version }

and try to `pod install` with original Podfile configuration 
(`pod 'FirebaseUI', :path => "../../FirebaseUI.podspec"`)

The result is...it's success.

I think that use `pod 'FirebaseUI'` in example/Podfile is more clear for many users who want to know how to work FirebaseUI.

dakeshi:
I'll try to fix this issues with new authentication. Should I keep the structure of the example source? I just show sign in button in the FirstViewController and call authViewController when clicked the button.

[...]

PR #78, opened 2016-07-04
Create the new example

### LINKED ISSUE #75: Unable to install FirebaseUI in the example
In addition to filename issue(#67),  FirebaseUIChat example has more issues about Podfile.
When I install the FirebaseUIChat with the following instructions in README, FirebaseUI pod is not existed in the workspace. But I get the installation success message from pod install.

- intallation
> cd FirebaseUI-iOS/examples/FirebaseUIChat
pod install
open FirebaseUIChat.xcworkspace

- pod install messages
> Installing Bolts (1.7.0)
Installing FBSDKCoreKit (4.13.1)
Installing FBSDKLoginKit (4.13.1)
Installing Firebase (3.2.1)
Installing FirebaseAnalytics (3.2.0)
Installing FirebaseAuth (3.0.2)
Installing FirebaseDatabase (3.0.1)
Installing FirebaseInstanceID (1.0.6)
**Installing FirebaseUI (0.4.0)**
Installing GoogleAppUtilities (1.1.1)
Installing GoogleAuthUtilities (2.0.1)
Installing GoogleInterchangeUtilities (1.2.1)
Installing GoogleNetworkingUtilities (1.2.1)
Installing GoogleSignIn (4.0.0)
Installing GoogleSymbolUtilities (1.1.1)
Installing GoogleUtilities (1.3.1)
Pod installation complete! There is 1 dependency from the Podfile and 16 total
pods installed.

If I edit Podfile with `pod 'FirebaseUI'` instead of `pod 'FirebaseUI', :path => "../../FirebaseUI.podspec"`,
FirebaseUI is successfully installed in the workspace. 

Is there some problem in FirebaseUI.podspec?

dakeshi:
I make PR to solve #75, #67 issues. 

# Add podspec for local dev
- Local pod is useful option to develop and test the project.
- Add podspec file in the pod folder.(https://guides.cocoapods.org/syntax/podfile.html#pod)
- Use git repo instead of downloading source files. 

>   s.source       = { :git => 'https://github.com/firebase/FirebaseUI-iOS.git', :tag => s.version }

- Resolve #75 issue

>   s.subspec 'AuthBase' do |authbase|
    authbase.source_files = 'Auth/AuthUI/Source/*.{h,m}'
    authbase.resource_bundles = {'FirebaseAuthUIBundle' => ['Auth/AuthUI/Resources/*.png', 'Auth/AuthUI/Strings/**/*.strings', 'Auth/AuthUI/Source/*.xib']}

# Create new example project
- simple google auth UI.
- The problem : If user try to sign in again, it still needs the authentication process. 
- How to test: Add GoogleService-Info.plist and URL scheme about REVERSED_CLIENT_ID


examples/FirebaseUIChat/FirebaseUIChat.xcodeproj/project.pbxproj
```diff
@@ -7,19 +7,19 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		C887CB2F78EAA58DC354FAFB /* libPods-FirebaseUIChat.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 585EBE168DE871301F5FA598 /* libPods-FirebaseUIChat.a */; };
```
mcdonamp: Please run `pod deintegrate` to get rid of the CoocaPods artifacts in the project.

mcdonamp:
Thanks for the PR! I'd prefer we not delete the Database example, so we should either show all of the components in a single example, or we should create two examples (one for Auth, one for DB). Eventually we'll have more modules (Storage, etc.), so I have some concerns about creating *n* examples, though on the flip side, creating and maintaining a mega example may be pretty difficult.

For now, I'd prefer we revert the deletions of the DB code and go for a single example, but I'm open to a discussion on this point.

examples/FirebaseUIChat/FirebaseUIChat.xcodeproj/project.pbxproj
dakeshi: Nice catch. 👍  fix it : abbd453

dakeshi:
> Thanks for the PR!

It's my pleasure. Thanks.

> For now, I'd prefer we revert the deletions of the DB code and go for a single example, but I'm open to a discussion on this point. 

I tried to keep the example structure as much as possible.
But there are two reasons to create the new example focused on the authUI. 

First, it is more natural for the user context. 
In my case, I find out FirebaseUI in the [Firebase Authentication documentation page](https://firebase.google.com/docs/auth/#key_functions)
It looks like FirebaseUI is a kind of authentication UI library according to the description in key functions section. If some users select their preferred platform, such as iOS, 
and move to the project page, what they want to see? AuthUI.
If an example shows auth and database in the same time, 
users need to see more documents to understand the example source.
So, I think that it is better to separate examples for auth and database if you want to keep database source.

Second, there are already a good example code for databaseUI in [quickstart-ios/database](https://github.com/firebase/quickstart-ios)

mcdonamp:
Ok, thanks for the rationale for the decision. Let me talk with some folks on the team--it's likely that we'll want to update quickstart-ios/auth to include FirebaseUI/Auth (since if that's what we're doing in the DB, it would make sense to do this with Auth), in which case this sample should include all FirebaseUI components, and we tell people to try the quickstarts for individual pieces.

mcdonamp:
Talked with the team (@puf, @samtstern, @abehaskins) and we think that the best way to present FirebaseUI will be to:
* Keep the example app here a "Catalog" type app (similar to [UIKitCatalog](https://developer.apple.com/library/ios/samplecode/UICatalog/Introduction/Intro.html) and what [our Android example app](https://github.com/firebase/FirebaseUI-Android#sample-app) is)
* Use the local source rather than shipped dependencies for the example app, to aid in local development
* Have the quickstart apps use FirebaseUI where appropriate
* Update the docs in this repo to add detailed DB documentation, so Auth isn't the only thing [here](https://github.com/firebase/FirebaseUI-iOS/tree/master/FirebaseUI).

mcdonamp:
Updated docs in #84

dakeshi:
Great feedback. Thank you and your team to have an interest about this issue 👍 

> Use the local source rather than shipped dependencies for the example app, to aid in local development

Nice approach. I think that my podspec can be a reference for this issue even though it is based on CocoaPods 1.0.0 not 0.39(#79) 

> Have the quickstart apps use FirebaseUI where appropriate

Good news. 💯 

> Update the docs in this repo to add detailed DB documentation, so Auth isn't the only thing

In addtion to database docs in this repo, I think that it is a good idea to mention FirebaseUI for database in [database intro page](https://firebase.google.com/docs/database/) like [auth docs does](https://firebase.google.com/docs/auth/) 

So, how should I do this PR? Close or reverting original source? Because you'll make the new example, reverting old source code would be useless and my podspec need to be test of its syntax on 0.39. Thoughts?

dakeshi:
refer to #89, #95

PR #80, opened 2016-07-05
Fix documentation

ppaulojr:
It's now `signInProviders` instead of `providers`

**Note**: CLA already signed.


mcdonamp:
I believe this is the same issue as #70, though @chaogoog can confirm.

chaogoog:
Yes, this is the same issue. I guess it doesn't hurt to fix the README now and then update it to `providers` in the next release.

PR #89, opened 2016-07-25
create sample chat app

morganchen12:
Hoping achieve sample parity with FirebaseUI-Android soon.

This code is going to serve as a reference for every developer who tries to use this library, so I'm not quite happy with where it is yet (it could use more comments, for starters)--but I think it's far along enough for some code review.

### DIFF
--- a/samples/swift/README.md
+++ b/samples/swift/README.md
@@ -4,7 +4,6 @@ Swift Samples
 This directory contains a collection of swift code samples. 
 
 In order to run the project you'll need a valid app in Firebase and 
-the `GoogleService-Info.plist` file for that project. Add the plist
-to the `uidemo` folder (the one with all the class files) and the project
-should build correctly. 
+the `GoogleService-Info.plist` file for that project. Drag the plist into the 
+project root and the project should build correctly. 
 
--- a/samples/swift/README.md
+++ b/samples/swift/README.md
@@ -1,0 +1,10 @@
+Swift Samples
+-----
+
+This directory contains a collection of swift code samples. 
+
+In order to run the project you'll need a valid app in Firebase and 
+the `GoogleService-Info.plist` file for that project. Add the plist
+to the `uidemo` folder (the one with all the class files) and the project
+should build correctly. 
+
--- a/samples/swift/README.md
+++ b/samples/swift/README.md
@@ -5,5 +5,18 @@ This directory contains a collection of swift code samples.
 
 In order to run the project you'll need a valid app in Firebase and 
 the `GoogleService-Info.plist` file for that project. Drag the plist into the 
-project root and the project should build correctly. 
+project root and the project should build correctly. Find more instructions
+and download a plist file from the [Firebase console](https://console.firebase.google.com). 
+
+###Chat Sample
+
+This sample uses [anonymous authentication](https://firebase.google.com/docs/auth/ios/anonymous-auth),
+so make sure anonymous auth is enabled in Firebase console.
+
+###Auth Sample
+
+This sample uses [email/password](https://firebase.google.com/docs/auth/ios/password-auth), 
+[Google](https://firebase.google.com/docs/auth/ios/google-signin), 
+and [Facebook](https://firebase.google.com/docs/auth/ios/facebook-login) 
+auth, so make sure those are enabled in Firebase console.
 
--- a/samples/swift/uidemo.xcodeproj/project.pbxproj
+++ b/samples/swift/uidemo.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		6BEBFBD2CFCDA8A7918028C1 /* Pods_uidemoTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 1A128E4DD412913B07E20C37 /* Pods_uidemoTests.framework */; };
+		8D16073E1D492B200069E4F5 /* AuthViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8D16073D1D492B200069E4F5 /* AuthViewController.swift */; };
 		8DABC9891D3D82D600453807 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC9881D3D82D600453807 /* AppDelegate.swift */; };
 		8DABC98B1D3D82D600453807 /* MenuViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC98A1D3D82D600453807 /* MenuViewController.swift */; };
 		8DABC98E1D3D82D600453807 /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 8DABC98C1D3D82D600453807 /* Main.storyboard */; };
@@ -37,6 +38,7 @@
 		5F8797FA94AF0055875F8117 /* Pods-uidemo.debug.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-uidemo.debug.xcconfig"; path = "Pods/Target Support Files/Pods-uidemo/Pods-uidemo.debug.xcconfig"; sourceTree = "<group>"; };
 		6EC625C321DD275A92C4085C /* Pods-uidemoTests.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-uidemoTests.release.xcconfig"; path = "Pods/Target Support Files/Pods-uidemoTests/Pods-uidemoTests.release.xcconfig"; sourceTree = "<group>"; };
 		8ADB97851843928A14BC9024 /* Pods-uidemo.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = "Pods-uidemo.release.xcconfig"; path = "Pods/Target Support Files/Pods-uidemo/Pods-uidemo.release.xcconfig"; sourceTree = "<group>"; };
+		8D16073D1D492B200069E4F5 /* AuthViewController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = AuthViewController.swift; sourceTree = "<group>"; };
 		8DABC9851D3D82D600453807 /* uidemo.app */ = {isa = PBXFileReference; explicitFileType = wrapper.application; includeInIndex = 0; path = uidemo.app; sourceTree = BUILT_PRODUCTS_DIR; };
 		8DABC9881D3D82D600453807 /* AppDelegate.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppDelegate.swift; sourceTree = "<group>"; };
 		8DABC98A1D3D82D600453807 /* MenuViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MenuViewController.swift; sourceTree = "<group>"; };
@@ -112,6 +114,7 @@
 				8DABC98A1D3D82D600453807 /* MenuViewController.swift */,
 				8DABC9AA1D3D947300453807 /* SampleCell.swift */,
 				8DABC9A81D3D872C00453807 /* Sample.swift */,
+				8D16073D1D492B200069E4F5 /* AuthViewController.swift */,
 				8D643F0F1D4827F10081F979 /* ChatSample */,
 				8DABC98C1D3D82D600453807 /* Main.storyboard */,
 				8DABC98F1D3D82D600453807 /* Assets.xcassets */,
@@ -355,6 +358,7 @@
 				8DABC98B1D3D82D600453807 /* MenuViewController.swift in Sources */,
 				8DABC9AB1D3D947300453807 /* SampleCell.swift in Sources */,
 				8DABC9A91D3D872C00453807 /* Sample.swift in Sources */,
+				8D16073E1D492B200069E4F5 /* AuthViewController.swift in Sources */,
 				8DABC9AD1D3D9EAF00453807 /* ChatViewController.swift in Sources */,
 				8DDF1AE51D3FF67D001F1160 /* ChatCollectionViewCell.swift in Sources */,
 				8DABC9891D3D82D600453807 /* AppDelegate.swift in Sources */,
@@ -492,7 +496,7 @@
 				CLANG_ENABLE_MODULES = YES;
 				INFOPLIST_FILE = uidemo/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
-				PRODUCT_BUNDLE_IDENTIFIER = com.firebase.uidemo;
+				PRODUCT_BUNDLE_IDENTIFIER = com.google.firebase.firebaseui.uidemo;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_OBJC_BRIDGING_HEADER = "";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
@@ -507,7 +511,7 @@
 				CLANG_ENABLE_MODULES = YES;
 				INFOPLIST_FILE = uidemo/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
-				PRODUCT_BUNDLE_IDENTIFIER = com.firebase.uidemo;
+				PRODUCT_BUNDLE_IDENTIFIER = com.google.firebase.firebaseui.uidemo;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SWIFT_OBJC_BRIDGING_HEADER = "";
 			};
--- a/samples/swift/uidemo.xcodeproj/project.pbxproj
+++ b/samples/swift/uidemo.xcodeproj/project.pbxproj
@@ -7,7 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		6BEBFBD2CFCDA8A7918028C1 /* Pods_uidemoTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 1A128E4DD412913B07E20C37 /* Pods_uidemoTests.framework */; };
 		8D16073E1D492B200069E4F5 /* AuthViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8D16073D1D492B200069E4F5 /* AuthViewController.swift */; };
 		8DABC9891D3D82D600453807 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC9881D3D82D600453807 /* AppDelegate.swift */; };
 		8DABC98B1D3D82D600453807 /* MenuViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC98A1D3D82D600453807 /* MenuViewController.swift */; };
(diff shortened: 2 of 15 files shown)


samtstern:
@iulukaya @silvolu FYI

samples/swift/uidemo.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,575 @@
...
+		};
+		8DABC9A31D3D82D600453807 /* Debug */ = {
+			isa = XCBuildConfiguration;
+			baseConfigurationReference = 5F8797FA94AF0055875F8117 /* Pods-uidemo.debug.xcconfig */;
+			buildSettings = {
+				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
+				CLANG_ENABLE_MODULES = YES;
+				INFOPLIST_FILE = uidemo/Info.plist;
+				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
+				PRODUCT_BUNDLE_IDENTIFIER = com.firebase.uidemo;
```
iulukaya: make the bundle something like "com.google.firebase.firebaseui.uidemo" to be more descriptive, uideo is no t a Firebase product but a demo under FirebaseUI project

samples/swift/uidemo.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,575 @@
...
+		8DABC9901D3D82D600453807 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 8DABC98F1D3D82D600453807 /* Assets.xcassets */; };
+		8DABC9931D3D82D600453807 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 8DABC9911D3D82D600453807 /* LaunchScreen.storyboard */; };
+		8DABC99E1D3D82D600453807 /* uidemoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC99D1D3D82D600453807 /* uidemoTests.swift */; };
+		8DABC9A91D3D872C00453807 /* Sample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC9A81D3D872C00453807 /* Sample.swift */; };
+		8DABC9AB1D3D947300453807 /* SampleCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC9AA1D3D947300453807 /* SampleCell.swift */; };
+		8DABC9AD1D3D9EAF00453807 /* ChatViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8DABC9AC1D3D9EAF00453807 /* ChatViewController.swift */; };
+		8DABC9AF1D3DA30D00453807 /* GoogleService-Info.plist in Resources */ = {isa = PBXBuildFile; fileRef = 8DABC9AE1D3DA30D00453807 /* GoogleService-Info.plist */; };
```
iulukaya: Delete this reference so it doesn't give error, but you can add a README like this, so that it tells how to add that file.

https://github.com/firebase/quickstart-ios/blob/master/invites/README.md

morganchen12: I'm not sure if I agree with this--the error is meaningful, and requiring every sample contributor to delete a reference in their project file before committing seems like a big hassle. Adding the README would be a good idea regardless, though.

ianbarber: I would take it out - I've seen more developers copy the file in again and end up with a broken reference and filled reference, as most just drag the plist into XCode. The error is meaningful, but also comes with a stack trace that has confused a decent number of people. We also don't include it in other samples - it should be all or nothing for consistency on our developer materials (though always happy to discuss changing them all if the other way makes more sense - but lets separate that conversation from this PR).

samples/swift/uidemo.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,579 @@
...
+			isa = PBXContainerItemProxy;
+			containerPortal = 8DABC97D1D3D82D600453807 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 8DABC9841D3D82D600453807;
+			remoteInfo = uidemo;
+		};
+/* End PBXContainerItemProxy section */
+
+/* Begin PBXFileReference section */
+		1A128E4DD412913B07E20C37 /* Pods_uidemoTests.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; includeInIndex = 0; path = Pods_uidemoTests.framework; sourceTree = BUILT_PRODUCTS_DIR; };
```
ianbarber: Deintegrate before checking in, else we will have issues when pod versions change.

samples/swift/uidemo/AuthViewController.swift
```diff
@@ -0,0 +1,15 @@
+//
```
ianbarber: Fix licensing header.

samples/swift/uidemo/ChatCollectionViewCell.swift
```diff
@@ -0,0 +1,75 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+
+class ChatCollectionViewCell: UICollectionViewCell {
```
ianbarber: Add a class comment describing purpose and intention of class for anything more than a few lines.

samples/swift/uidemo/ChatCollectionViewCell.swift
```diff
@@ -0,0 +1,75 @@
...
+  }
+  
+  @IBOutlet var containerView: UIView! {
+    didSet {
+      containerView.layer.cornerRadius = 8
+      containerView.layer.masksToBounds = true
+    }
+  }
+  
+  // These constraints are used to left- and right-align chat bubbles.
```
ianbarber: Make this a proper doc comment

samples/swift/uidemo/ChatCollectionViewCell.swift
```diff
@@ -0,0 +1,75 @@
...
+      leadingConstraint.identifier = "leading constraint"
+    }
+  }
+  @IBOutlet private(set) var trailingConstraint: NSLayoutConstraint! {
+    didSet {
+      trailingConstraint.identifier = "trailing constraint"
+    }
+  }
+  
+  // This is the source of truth for the message font,
```
ianbarber: Ditto

samples/swift/uidemo/ChatCollectionViewCell.swift
```diff
@@ -0,0 +1,75 @@
...
+  // overriding whatever is set in interface builder.
+  static var messageFont: UIFont {
+    return UIFont.systemFontOfSize(UIFont.systemFontSize())
+  }
+  
+  // Colors for messages sent by the client.
+  static var selfColors: (background: UIColor, text: UIColor) {
+    return (
+      background: UIColor(red: 21 / 255, green: 60 / 255, blue: 235 / 255, alpha: 1),
```
ianbarber: Feels like these colours should be defined somewhere centralish so they can be reused.

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+    guard let uid  = dict["uid"]  else { return nil }
+    guard let text = dict["text"] else { return nil }
+    
+    self.name = name
+    self.uid = uid
+    self.text = text
+  }
+}
+
+// View controller demonstrating using a FirebaseCollectionViewDataSource
```
ianbarber: Format function commentproperly (see http://nshipster.com/swift-documentation/ for examples)

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+// to populate a collection view with chat messages. The relevant code
+// is in the call to `collectionViewDataSource.populateCellWithBlock`.
+//
+// All of the error handling in this controller is done with `fatalError`;
```
ianbarber: Hah! People definitely will - can you move/replicate this comment to close to sections users are likely to cut and paste, or when errors will appear.

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+  
+  override func viewDidAppear(animated: Bool) {
+    super.viewDidAppear(animated)
+    
+    self.auth?.signInAnonymouslyWithCompletion { (user, error) in
+      if let error = error {
+        fatalError("Sign in failed: \(error.localizedDescription)")
+      }
+      
+      self.query = self.chatReference.queryLimitedToLast(50)
```
ianbarber: Do auth related actions in an auth state listener rather than in a response to a sign in completition - this is (generally) a better pattern for real apps to adopt, but it does come with its own set of constraints (e.g. it must be somewhat idempotent as auth state changes can take place without in-app user action)

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+      
+      self.collectionViewDataSource = FirebaseCollectionViewDataSource(query: self.query!,
+        prototypeReuseIdentifier: ChatViewController.reuseIdentifier,
+        view: self.collectionView)
+      
+      // The initializer called above--though it takes a collection view--
```
ianbarber: This comment is good, but this block in general is pretty long and doing quite a lot of work. Consider adding some more narrative comments so that a reader skimming through the code to pull out examples will have an idea of the context they're copying from.

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+      // doesn't actually set the collection view's data source, so if
+      // we don't set it before trying to populate our view our app will crash.
+      self.collectionView.dataSource = self.collectionViewDataSource
+      
+      self.collectionViewDataSource.populateCellWithBlock { (anyCell, data) in
```
ianbarber: The collection view inner block is also quite long and kind of distracts. Consider moving this setup to a separate function and sticking it lower in the controller (as its largely app specific)

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+          let colors = ChatCollectionViewCell.othersColors
+          cell.containerView.backgroundColor = colors.background
+          cell.textLabel.textColor = colors.text
+          cell.leadingConstraint.active = false
+          cell.trailingConstraint.constant = constant
+          cell.trailingConstraint.active = true
+        }
+      }
+      
+      // Unfortunately neither FirebaseArray nor FirebaseCollectionViewDataSource
```
ianbarber: If its "unfortunately" and you work on it - i presume there is a tracking bug open for adding said listeners :)

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+                                                     object: nil)
+  }
+  
+  override func viewWillDisappear(animated: Bool) {
+    super.viewWillDisappear(animated)
+    NSNotificationCenter.defaultCenter().removeObserver(self)
+  }
+  
+  deinit {
+    NSNotificationCenter.defaultCenter().removeObserver(self)
```
ianbarber: Can we get a deinit without a viewWillDisappear?

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+  }
+  
+  @objc private func keyboardWillShow(notification: NSNotification) {
```
ianbarber: This is boilerplatey. For sample code where we expect code readers to be higher than normal its worth organizing the code so the demonstrative stuff is higher up, and the generic stuff lower down (within bounds of normal language conventions).

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+    UIView.setAnimationCurve(curve)
+    UIView.animateWithDuration(duration) {
+      self.view.layoutIfNeeded()
+    }
+  }
+  
+  @objc private func didTapSend(sender: AnyObject) {
+    guard let user = self.auth?.currentUser else { return }
+    let uid = user.uid
+    let name = "User " + (uid as NSString).substringToIndex(6)
```
ianbarber: Is casting to NSString really required to substring?

samples/swift/uidemo/ChatViewController.swift
```diff
@@ -0,0 +1,251 @@
...
+      if let error = error {
+        fatalError("Failed to write message: \(error.localizedDescription)")
+      }
+    }
+    
+    self.textView.text = ""
+  }
+  
+  private func scrollToBottom(animated animated: Bool) {
+    // FirebaseArray's documentation online is pretty lacking.
```
ianbarber: Bug pls.

samples/swift/uidemo/MenuViewController.swift
```diff
@@ -0,0 +1,63 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+
+// This controller exists solely to list the samples we've defined thus far.
```
ianbarber: Useful comment :+1:

samples/swift/uidemo/Sample.swift
```diff
@@ -0,0 +1,58 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+
+// As we add more sample use cases to FirebaseUI,
```
ianbarber: This describes the intent, but in a roundabout way. Describe what it is, and then how it should be used.

samples/swift/uidemo/SampleCell.swift
```diff
@@ -0,0 +1,48 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+
+class SampleCell: UITableViewCell {
```
ianbarber: As an example, this class is small and straightforward enough I wouldn't feel the need for a class comment unless you were being complete.

samples/swift/uidemo/ChatViewController.swift
morganchen12: This is the lazy (bad) way of getting substrings by disregarding Unicode. Swift's substring thing is much more verbose for Unicode reasons.

```swift
let other = str[str.characters.startIndex..<str.characters.startIndex.advancedBy(6)]
```

This code behaves subtly differently than the code in the sample, and I have no preference as to which one we keep. Should I replace the NSString cast with the Swift string subscript?

samples/swift/uidemo/ChatViewController.swift
morganchen12: It is, I was using the code comments to take notes to myself and left this here. Will clean up :+1:

samples/swift/uidemo.xcodeproj/project.pbxproj
morganchen12: I think we're talking about different errors. When I try running this with the reference but without the plist, I get a compiler error saying it can't find the file. Dragging the plist into Xcode then produces two valid references pointing to the same file (ugly, would like to avoid), but the project does build and run fine. Removing the reference gives me a runtime error on app launch that says it can't find the plist file. In this case I still would prefer the former, since the red broken reference in Xcode gives some more context as to where the project expects the plist file to be; the runtime error says it can't find it, but doesn't mention where it looked.

The distinction is small imo and not a killer either way and seems largely dependent on Xcode behavior (which I'm willing to disregard for the sake of consistency), but I wanted to disambiguate stuff before making changes.

samples/swift/uidemo.xcodeproj/project.pbxproj
morganchen12: what's deintegrate?

ianbarber: pod deintegrate - it is a plugin (bundled from 1.0) that removes all of the cocoapods specific stuff from the pbxproj. Shouldn't chance the basic flow (pod update/install then go), but means you don't hit silliness when pod install is run again a project that already has pod stuff from an older version (e.g. if the names of build phases change due to emoji boxes being added or similar, you end up with two sets of build phases and sadness ensues)

samples/swift/uidemo/ChatViewController.swift
ianbarber: That's interesting, thanks for the background! That range thing is pretty ugly, but I think we should go with it over the cast.

samples/swift/uidemo/ChatViewController.swift
ianbarber: Fair play! A reasonable convention in those cases is todo(username):
// TODO(morganchen): Defrobnicate this. 

I'm not totally sold by that type of comment, but at least its clear (for when you come back to look at a bug two years later and realise it has a comment saying "really important to fix this before launch" 🐩 )

samples/swift/uidemo.xcodeproj/project.pbxproj
ianbarber: Hmm, interesting - thats a reasonable point that both refs would be filled. I would say that for conventions sake I would remove it here. However, I think you have a decent argument. Maybe track this in a bug around moving all sample code to include the plist if you think the net experience is better, and we can have the discussion there (and do the change en masse as necessary)

dakeshi: vote 1 to delete GoogleService-Info.plist reference 
because we can get information where the file should be located, how to do it from several references:
https://firebase.google.com/docs/ios/setup
https://www.youtube.com/user/Firebase

Without a reference, users got the following errors.
>2016-07-28 11:19:28.042 uidemo[2728:68013] Could not locate configuration file: 'GoogleService-Info.plist'.
2016-07-28 11:19:28.087 uidemo[2728:68013] *** Terminating app due to uncaught exception 'com.firebase.core', reason: '[FIRApp configure] could not find a valid GoogleServices-Info.plist in your project. Please download one from https://console.firebase.google.com/.

So, if you are worried about the location of the plist file, make enhanced error message would be one solution. It would be looks like this:

[FIRApp configure] could not find a valid GoogleServices-Info.plist in your project. Please download one from https://console.firebase.google.com/. **You can get more information in https://firebase.google.com/docs/ios/setup and https://www.youtube.com/user/Firebase**

morganchen12: Reference removed :+1:

The `FIRApp.configure()` exception is one that comes from inside Firebase and not FirebaseUI, so I can't change it with this PR. It is worth noting though.

morganchen12:
Code should be ready for another round of review.

samples/swift/uidemo.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,539 @@
...
+			sourceTree = "<group>";
+		};
+/* End PBXGroup section */
+
+/* Begin PBXNativeTarget section */
+		8DABC9841D3D82D600453807 /* uidemo */ = {
+			isa = PBXNativeTarget;
+			buildConfigurationList = 8DABC9A21D3D82D600453807 /* Build configuration list for PBXNativeTarget "uidemo" */;
+			buildPhases = (
+				757DE2C28DE66EF8748CA3DA /* [CP] Check Pods Manifest.lock */,
```
ianbarber: DId these get left behind even after a pod deintegrate? Interesting if so.

samples/swift/README.md
```diff
@@ -0,0 +1,9 @@
+Swift Samples
+-----
+
+This directory contains a collection of swift code samples. 
+
+In order to run the project you'll need a valid app in Firebase and
```
ianbarber: Include a little bit more setup instructions (e.g. go to console.firebase.google.com, setup a app with bundle id <whatever it is>, enable any auth providers required etc. )

ianbarber:
Couple of nits, but otherwise LGTM.

samples/swift/uidemo.xcodeproj/project.pbxproj
morganchen12: Yeah, seems like it.

morganchen12:
Should be good to merge now.

Issue #92, opened 2016-07-27
Add Facebook scope customization


samtstern:
Context: https://github.com/firebase/FirebaseUI-Android/pull/224

Users should be able to pass a custom array of scopes for Facebook sign in using FirebaseUI.  On Android this can be achieved simply by providing a string resource to be overridden.

morganchen12:
FBUI-iOS already does allow this via a mutable array property on `FIRFacebookAuthUI`, so you can go ahead and accept that PR without worrying about feature parity.

I doubt we'll end up doing the same resource file thing though since for whatever reason it's not popular on iOS to shove everything in a plist (especially with Swift, where you lose type information when loading stuff from plists instead of just sticking it in code).

The pull request I made just adds this mutable property to the designated initializer, making it even more designated than before.

samtstern:
Fix implemented in #94, will go into the next release.

PR #94, opened 2016-07-29
make designated initializer include fb app permissions

morganchen12:
Addresses #92.


samtstern:
LGTM, thanks for doing this!

PR #95, opened 2016-08-01
Auth sample

morganchen12:
Adds the swift auth sample to the catalogue app.

### DIFF
--- a/samples/swift/uidemo/AppDelegate.swift
+++ b/samples/swift/uidemo/AppDelegate.swift
@@ -34,8 +34,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     return true
   }
   
-  func application(application: UIApplication, openURL url: NSURL, sourceApplication: String?, annotation: AnyObject) -> Bool {
-    // Seems like an oversight that `sourceApplication` should be force-unwrapped.
+  func application(app: UIApplication, openURL url: NSURL, options: [String : AnyObject]) -> Bool {
+    let sourceApplication = options[UIApplicationOpenURLOptionsSourceApplicationKey] as! String?
+    // Seems like an oversight that this API doesn't take a nullable sourceApplication.
     if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication!) ?? false {
       return true
     }
--- a/samples/swift/uidemo/AppDelegate.swift
+++ b/samples/swift/uidemo/AppDelegate.swift
@@ -16,6 +16,7 @@
 
 import UIKit
 import Firebase
+import FirebaseAuthUI
 
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
@@ -31,6 +32,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     // accompanying valid GoogleService-Info.plist file.
     FIRApp.configure()
     return true
+  }
+  
+  func application(application: UIApplication, openURL url: NSURL, sourceApplication: String?, annotation: AnyObject) -> Bool {
+    // Seems like an oversight that `sourceApplication` should be force-unwrapped.
+    if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication!) ?? false {
+      return true
+    }
+    // other URL handling goes here.
+    return false
   }
 }
 
--- a/samples/swift/uidemo/AppDelegate.swift
+++ b/samples/swift/uidemo/AppDelegate.swift
@@ -36,8 +36,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
   
   func application(app: UIApplication, openURL url: NSURL, options: [String : AnyObject]) -> Bool {
     let sourceApplication = options[UIApplicationOpenURLOptionsSourceApplicationKey] as! String?
-    // Seems like an oversight that this API doesn't take a nullable sourceApplication.
-    if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication!) ?? false {
+    if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication) ?? false {
       return true
     }
     // other URL handling goes here.
--- a/samples/swift/uidemo/AuthViewController.swift
+++ b/samples/swift/uidemo/AuthViewController.swift
@@ -47,6 +47,13 @@ class AuthViewController: UIViewController {
   @IBOutlet private var signOutButton: UIButton!
   @IBOutlet private var startButton: UIButton!
   
+  @IBOutlet private var signedInLabel: UILabel!
+  @IBOutlet private var nameLabel: UILabel!
+  @IBOutlet private var emailLabel: UILabel!
+  @IBOutlet private var uidLabel: UILabel!
+  
+  @IBOutlet var topConstraint: NSLayoutConstraint!
+  
   override func viewWillAppear(animated: Bool) {
     super.viewWillAppear(animated)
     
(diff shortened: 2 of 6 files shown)


morganchen12:
This sample is considerably more gross than the last one, since each individual login provider requires their own sort of app setup with instructions on their own websites and no one wants to check their API keys or OAuth tokens into a public repo.

samples/swift/uidemo/AuthViewController.swift
```diff
@@ -15,8 +15,116 @@
 //
 
 import UIKit
+import Firebase
+import FirebaseAuthUI
+import FirebaseGoogleAuthUI
+import FirebaseFacebookAuthUI
 
+let kFirebaseTermsOfService = NSURL(string: "https://www.firebase.com/terms/terms-of-service.html")!
```
iulukaya: https://firebase.google.com/terms/

samples/swift/uidemo/Info.plist
```diff
@@ -18,6 +18,17 @@
...
 	<key>CFBundleSignature</key>
 	<string>????</string>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>com.googleusercontent.apps.whatever</string>
```
iulukaya: <string>REVERSED_CLIENT_ID</string>

samples/swift/uidemo/AppDelegate.swift
```diff
@@ -32,5 +33,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     FIRApp.configure()
     return true
   }
+  
+  func application(app: UIApplication, openURL url: NSURL, options: [String : AnyObject]) -> Bool {
+    let sourceApplication = options[UIApplicationOpenURLOptionsSourceApplicationKey] as! String?
+    // Seems like an oversight that this API doesn't take a nullable sourceApplication.
```
ianbarber: Raise issue.

samples/swift/uidemo/AuthViewController.swift
```diff
@@ -15,8 +15,116 @@
...
+    // will still appear in the UI, but they'll crash the app when tapped.
+    let providers: [FIRAuthProviderUI] = [
+      FIRGoogleAuthUI(clientID: kGoogleAppClientID)!,
+      FIRFacebookAuthUI(appID: kFacebookAppID)!,
+    ]
+    self.authUI?.signInProviders = providers
+    
+    // Strangely this is listed as TOSURL in the objc source and isn't
+    // given a swift name that would otherwise make it import as termsOfServiceURL.
+    self.authUI?.termsOfServiceURL = kFirebaseTermsOfService
```
ianbarber: Definitely resolve this.

samples/swift/uidemo/AuthViewController.swift
```diff
@@ -15,8 +15,116 @@
...
+    super.viewWillDisappear(animated)
+    if let handle = self.authStateDidChangeHandle {
+      self.auth?.removeAuthStateDidChangeListener(handle)
+    }
+  }
+  
+  @IBAction func startPressed(sender: AnyObject) {
+    // The function signature says it returns a view controller,
+    // but when called it actually returns a closure returning a view controller.
+    // Maybe this is a swift-objc interoperability bug.
```
ianbarber: Sounds very plausible. If there is a fix we can do to avoid it though would be good - either track an issue (and doc in header for that method) or fix if you can find one!

samples/swift/uidemo/AuthViewController.swift
```diff
@@ -15,8 +15,116 @@
...
 
+let kFirebaseTermsOfService = NSURL(string: "https://firebase.google.com/terms/")!
+
+// Your Google app's client ID, which can be found in the GoogleService-Info.plist file
+// and is stored in the `clientID` property of your FIRApp options.
+// Firebase Google auth is built on top of Google sign-in, so you'll have to add a URL
+// scheme to your project as outlined at the bottom of this reference:
+// https://developers.google.com/identity/sign-in/ios/start-integrating
+//
+// Make sure you don't accidentally check in your client ID in a public repo!
```
ianbarber: You absolutely can, its basically not secret

ianbarber:
This LGTM - I'm happy for you to merge at your discretion on this one, and we can do follow ups in separate reqs - it may be easier to investigate that terms of service issue with it in the repo.

samples/swift/uidemo/AppDelegate.swift
```diff
@@ -32,5 +33,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication!) ?? false {
```
dakeshi: `sourceApplication` is String optional. So, I think that `sourceApplication!` could be nil and make a runtime errors. Isn't it?

how was this?
```swift
    if let sourceApplication = options[UIApplicationOpenURLOptionsSourceApplicationKey] as! String? {
        // Seems like an oversight that this API doesn't take a nullable sourceApplication.
        if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication) ?? false {
```

samples/swift/uidemo/Info.plist
```diff
@@ -18,6 +18,17 @@
...
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>REVERSED_CLIENT_ID</string>
+			</array>
+		</dict>
```
dakeshi: Facebook login needs custom URL schemes and LSApplicationQueriesSchemes.
In addtion to [Auth README](https://github.com/firebase/FirebaseUI-iOS/tree/master/FirebaseUI/Auth),
it would be helpful to provide the default settings. 

it looks like the following:

	<array>
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Editor</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>REVERSED_CLIENT_ID</string>
			</array>
		</dict>
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Editor</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string>FACEBOOK_APP_ID</string>
			</array>
		</dict>
	</array>
    <key>LSApplicationQueriesSchemes</key>
    <array>
        <string>fbapi</string>
        <string>fb-messenger-api</string>
        <string>fbauth2</string>
        <string>fbshareextension</string>
    </array>

samples/swift/uidemo/AuthViewController.swift
dakeshi: Um.. Why did you use `FIRAuthUI.authViewController(self.authUI!)()` instead of `self.authUI!.authViewController()`?
authViewController() is an instance method.

samples/swift/uidemo/AuthViewController.swift
iulukaya: It's termsOfServiceURL in pod version of objc source too. Most likely pod version and github versions are not the same.

iulukaya:
LGTM after addressing few comments

samples/swift/uidemo/AuthViewController.swift
ianbarber: Good find! Yeah, we should fix this then, and make sure we're testing against the source version.

samples/swift/uidemo/AuthViewController.swift
morganchen12: Yeah, I'm not sure how this code ever compiled.

samples/swift/uidemo/Info.plist
morganchen12: added, thanks!

samples/swift/uidemo/AuthViewController.swift
morganchen12: Do we want the sample to be against the source version? Pod version is going to be what everyone encounters as they start using the lib.

samples/swift/uidemo/AuthViewController.swift
dakeshi: Yes. It's a very interesting that it doesn't make a error.

coyistik:
@iulukaya  Hayvan gibi Git kullanıyorsunuz tebrik ederim :)  :+1:

PR #100, opened 2016-08-02
fix nullability annotations in authUI

### LINKED ISSUE #99: AuthUI handleOpenURL(url:sourceApplication:) should take nullable sourceApplication
Otherwise there's this awful disconnect between the app delegate callback method which sometimes gives a `nil` sourceApplication and the AuthUI interface which demands a nonnull sourceApplication.

morganchen12:
Fixes #99.

### DIFF
--- a/FirebaseUI/Auth/AuthUI/Source/FIRAuthProviderUI.h
+++ b/FirebaseUI/Auth/AuthUI/Source/FIRAuthProviderUI.h
@@ -99,7 +99,7 @@ typedef void (^FIRAuthProviderSignInCompletionBlock)(FIRAuthCredential *_Nullabl
     @param sourceApplication The application which tried opening the URL.
     @return YES if your auth provider handled the URL. NO otherwise.
  */
-- (BOOL)handleOpenURL:(NSURL *)URL sourceApplication:(NSString *)sourceApplication;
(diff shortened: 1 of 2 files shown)


mcdonamp:
Mostly since I'm curious...

What's the difference between:
```
- fooWithBar:(nullable Bar *)bar;
```

And:
```
- fooWithBar:(Bar *_Nullable)bar;
```

Looks like nothing: http://stackoverflow.com/questions/32452889/difference-between-nullable-nullable-and-nullable-in-objective-c

Also, looks like the [Google Objective-C style guide](https://google.github.io/styleguide/objcguide.xml) is missing guidance on nullability.

I prefer the first (and would like we use them everywhere we're allowed to, aka everywhere the compiler doesn't throw a fit [typedefs and blocks have special rules]).

morganchen12:
They're the same. There's also a third:

```objc
- (void)fooWithBar:(Bar *__nullable)bar;
```

The postfix `_Nullable` can be more useful(?) when referring to references of references, like:

```objc
- (void)fooWithBar:(Bar *)bar error:(NSError *_Nullable *_Nonnull)error;
```

but it's such a rare case that it probably doesn't matter. We should just pick one and commit to it (and update the Google style guide eventually). The prefix version definitely looks nicer.

mcdonamp:
Cool, thanks for the clarification :)

PR #105, opened 2016-08-04
fix the optional error with the optional binding

dakeshi:
optional `sourceApplication` has to use with the optional binding to prevent the optional syntax error. 

With the optional binding, `sourceApplication` which is the handleOpenURL method argument is always has a value.

Is there another solution?

### DIFF
--- a/samples/swift/uidemo/AppDelegate.swift
+++ b/samples/swift/uidemo/AppDelegate.swift
@@ -35,9 +35,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
   }
   
   func application(app: UIApplication, openURL url: NSURL, options: [String : AnyObject]) -> Bool {
-    let sourceApplication = options[UIApplicationOpenURLOptionsSourceApplicationKey] as! String?
-    if FIRAuthUI.authUI()?.handleOpenURL(url, sourceApplication: sourceApplication) ?? false {
-      return true
+    if let sourceApplication = options[UIApplicationOpenURLOptionsSourceApplicationKey] as! String? {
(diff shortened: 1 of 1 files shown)


morganchen12:
This code always returns `false` if there was no source application even if the app can handle the available URL, which is not desired behavior. The ugly workaround with `sourceApplication ?? ""` tries to resolve this by hoping every URL handler treats empty or unrecognized source applications as they treat `nil`, which is closer to but not necessarily the desired behavior. Ideally all of the URL handlers should know how to deal with `nil`.

We'll be able to eventually remove this since #100 was merged, but for now I'd prefer the `?? ""` workaround.

dakeshi:
Thank you for your feedback @morganchen12.

> This code always returns false if there was no source application even if the app can handle the available URL, which is not desired behavior.

It sounds like that handleOpenURL method of the authUI works well regardless sourceApplication argument value. If so, I think that handleOpenURL method api need to be updated to have the only one argument: the type of NSURL.

And then we'll have more simpler syntax:

```swift
    if FIRAuthUI.authUI()?.handleOpenURL(url) {
      return true
    }

    return false
...
```

  > but for now I'd prefer the ?? "" workaround.
Yes, it would be nicer solution temporally.

morganchen12:
`handleOpenURL` calls through to a protocol method on each of the login provider objects that the `FIRAuthUI` instance knows about, so having a `sourceApplication` could be useful to them (i.e. Facebook auth might expect that its callback came from the Facebook app or Safari and fail otherwise). But the optionality is definitely incorrect currently, and will be fixed in the next release.

dakeshi:
Fixed it : a5197f6

Remove the optional binding and use `sourceApplicaiton ?? ""`

morganchen12:
Thanks!

PR #107, opened 2016-08-04
Make APIs more idiomatic

morganchen12:
In endless pursuit of that native platform look-and-feel™.

This may break things.

But our pod distribution is currently a `vendored_framework` with a zip download, so ¯\\\_(ツ)\_/¯

### DIFF
--- a/FirebaseUI/Database/API/FirebaseArray.h
+++ b/FirebaseUI/Database/API/FirebaseArray.h
@@ -39,7 +39,7 @@ NS_ASSUME_NONNULL_BEGIN
  * The delegate object that array changes are surfaced to, which conforms to the
  * [FirebaseArrayDelegate Protocol](FirebaseArrayDelegate).
  */
-@property(weak, nonatomic) id<FirebaseArrayDelegate> delegate;
+@property(weak, nonatomic, nullable) id<FirebaseArrayDelegate> delegate;
 
 /**
  * The query on a Firebase reference that provides data to populate the instance of FirebaseArray.
@@ -50,6 +50,16 @@ NS_ASSUME_NONNULL_BEGIN
  * The delegate object that array changes are surfaced to.
  */
 @property(strong, nonatomic) NSMutableArray<FIRDataSnapshot *> * snapshots;
+
+/**
+ * The number of objects in the FirebaseArray.
+ */
+@property(nonatomic, readonly) NSUInteger count;
+
+/**
+ * The items currently in the FirebaseArray.
+ */
+@property(nonatomic, readonly, copy) NSArray *items;
 
 #pragma mark -
 #pragma mark Initializer methods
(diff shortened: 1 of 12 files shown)


FirebaseUI/Database/API/FirebaseArray.h
```diff
@@ -51,6 +51,16 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+/**
+ * The number of objects in the FirebaseArray.
+ */
+@property(nonatomic, readonly) NSUInteger count;
+
+/**
+ * The items currently in the FirebaseArray.
+ */
+@property(nonatomic, readonly, copy) NSArray *items;
```
morganchen12: Wanted to expose a plain immutable data type since `FirebaseArray` is full of behavior despite pretending to be an array.

FirebaseUI/Database/API/FirebaseArrayDelegate.h
```diff
@@ -36,7 +38,7 @@
  * @param object The object added to the FirebaseArray
  * @param index The index the child was added at
  */
-- (void)childAdded:(id)object atIndex:(NSUInteger)index;
+- (void)array:(FirebaseArray *)array didAddObject:(id)object atIndex:(NSUInteger)index;
```
morganchen12: Renamed to follow Cocoa convention; now has more clear semantic of when the call happens (immediate after insertion, as opposed to a hypothetical `willAddObject:` method that could be added in the future).

FirebaseUI/Database/Implementation/FirebaseDataSource.m
```diff
@@ -21,7 +21,14 @@
...
 
 @interface FirebaseDataSource ()
-@property(copy, nonatomic) void (^cancelBlock)(NSError *);
+
+@property(copy, nonatomic, nonnull) void (^cancelBlock)(NSError *);
+
+/**
+ * The FirebaseArray which backs the instance of the datasource.
+ */
+@property(strong, nonatomic, nonnull) FirebaseArray *array;
```
morganchen12: Made this private because it has all sorts of behaviors. Previously it was possible to accidentally break a `FirebaseDataSource` by assigning the delegate of the array to something else.

morganchen12:
@mcdonamp FYI

morganchen12:
Actually while I'm breaking stuff I might as well break everything

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,22 +20,28 @@
...
-}
-
 - (instancetype)initWithQuery:(FIRDatabaseQuery *)query {
+  NSParameterAssert(query != nil);
   self = [super init];
   if (self) {
-    self.snapshots = [NSMutableArray array];
-    self.query = query;
+    _snapshots = [NSMutableArray array];
+    _query = query;
```
iulukaya: why do we access to ivars directly, instead of using setters as in self.query =

morganchen12: Take a look at #108, that's where all the cool new dev is going on :)

mcdonamp:
Setters may have side effects while ivars won't, so always set ivars in `init`.

You can thank my horrific past self for this bad code.

PR #108, opened 2016-08-05
DatabaseUI Tests: Road to green CI

morganchen12:
The ultimate goal of this pull request is to make FirebaseUI more familiar for new contributors who are already used to the iOS OSS landscape. Ideally, any developer should be able to fork the repo and make changes with confidence that the unit tests will tell them if their changes have broken anything.

The first step of that is to have tests.


FirebaseUI.xcodeproj/project.pbxproj
```diff
@@ -1035,163 +1132,194 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
-		072ABD579ED6CE7C8FC98658 /* 📦 Check Pods Manifest.lock */ = {
+		072ABD579ED6CE7C8FC98658 /* [CP] Check Pods Manifest.lock */ = {
```
ianbarber: Deintegrate

FirebaseUI/Auth/AuthProviderUI/Facebook/Source/FIRFacebookAuthUI.h
```diff
@@ -34,26 +34,26 @@ NS_ASSUME_NONNULL_BEGIN
     @brief The scopes to use with Facebook Login.
     @remarks Defaults to using "email" scopes.
  */
-@property(nonatomic, copy) NSArray<NSString *> *scopes;
+@property(nonatomic, copy, readonly) NSArray<NSString *> *scopes;
```
ianbarber: Is this a behavior change, or just clarifying how it was before?

FirebaseUI/Auth/AuthProviderUI/Google/Source/FIRGoogleAuthUI.m
```diff
@@ -31,10 +31,15 @@
...
+/** @var kGooglePlusMeScope
+ @brief The OAuth scope string for the user's email scope.
+ */
+static NSString *const kGoogleUserInfoEmailScope = @"https://www.googleapis.com/auth/userinfo.email";
+
+/** @var kGooglePlusMeScope
+ @brief The OAuth scope string for the basic G+ profile information scope.
  */
-static NSString *const kGooglePlusScopesPrefix = @"https://www.googleapis.com/auth/plus.";
+static NSString *const kGoogleUserInfoProfileScope = @"https://www.googleapis.com/auth/userinfo.profile";
```
ianbarber: This one is a behavior change I think, though the effective difference between those scopes is slight, an upgrade may cause a reconsent. At the least we should flag this in changelog.

FirebaseUI/Auth/AuthUI/Source/FIRAuthUI.h
```diff
@@ -67,7 +67,7 @@ typedef void (^FIRAuthUIResultCallback)(FIRUser *_Nullable user, NSError *_Nulla
     @brief Gets the @c FIRAuthUI object for the default FirebaseApp.
     @remarks Thread safe.
  */
-+ (nullable FIRAuthUI *)authUI NS_SWIFT_NAME(authUI());
++ (nullable FIRAuthUI *)defaultAuthUI NS_SWIFT_NAME(authUI());
```
ianbarber: Idiomatic to have the swift name stay authUI?

FirebaseUI/Database/API/FirebaseArray.h
```diff
@@ -39,43 +49,36 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property(nonatomic, readonly) NSUInteger count;
 
 /**
- * Intitalizes FirebaseArray with a standard Firebase reference.
- * @param ref The Firebase reference which provides data to FirebaseArray
- * @return The instance of FirebaseArray
+ * The items currently in the FirebaseArray.
  */
-- (instancetype)initWithRef:(FIRDatabaseReference *)ref;
+@property(nonatomic, readonly, copy) NSArray *items;
```
ianbarber: Does copy make sense here with readonly? I notice you have the same set of params earlier as well, but in a different order: reorder one set. May be worth checking the Google objective C style guide to see if there is an ordering recommendation.

FirebaseUI/Database/API/FirebaseArrayDelegate.h
```diff
@@ -36,7 +38,7 @@
  * @param object The object added to the FirebaseArray
  * @param index The index the child was added at
  */
-- (void)childAdded:(id)object atIndex:(NSUInteger)index;
+- (void)array:(FirebaseArray *)array didAddObject:(id)object atIndex:(NSUInteger)index;
```
ianbarber: This change is pretty impactful I think, people do use FirebaseArray directly. We may want a migration cheat sheet - replace X with Y kind of thing.

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,113 +20,142 @@
 
 #import "FirebaseArray.h"
 
+@interface FirebaseArray ()
+
+/**
+ * The delegate object that array changes are surfaced to.
+ */
+@property(strong, nonatomic) NSMutableArray<FIRDataSnapshot *> * snapshots;
```
ianbarber: ws after *

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,113 +20,142 @@
 
 #import "FirebaseArray.h"
 
+@interface FirebaseArray ()
+
+/**
+ * The delegate object that array changes are surfaced to.
```
ianbarber: Comment doesn't match prop.

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,113 +20,142 @@
...
-#pragma mark -
-#pragma mark Initializer methods
-
-- (instancetype)initWithRef:(FIRDatabaseReference *)ref {
-  return [self initWithQuery:ref];
-}
+#pragma mark - Initializer methods
 
 - (instancetype)initWithQuery:(FIRDatabaseQuery *)query {
+  NSParameterAssert(query != nil);
```
ianbarber: This worries me slightly - would the behavior in the previous version have been a crash if a null query was passed anyway?

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,113 +20,142 @@
...
   self = [super init];
   if (self) {
-    self.snapshots = [NSMutableArray array];
-    self.query = query;
+    _snapshots = [NSMutableArray array];
```
ianbarber: Nit: I prefer assignment to use the self. keyword as then you know the property modifiers are respected (particularly for copy etc). Reads should use the ivar.

FirebaseUI/Database/API/FirebaseArrayDelegate.h
samtstern: FWIW we consider FirebaseArray to be a "private API" even though some are accessing it.  So we can flag it in the changelog but we don't have to worry too much.

On Android it's not possible to use FirebaseArray directly due to the access modifiers.

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,113 +20,142 @@
...
-        NSUInteger index = [self indexForKey:previousChildKey] + 1;
+        NSUInteger index = 0;
+        if (previousChildKey != nil) {
+          index = [self indexForKey:previousChildKey] + 1;
+        }
 
         [self.snapshots insertObject:snapshot atIndex:index];
 
-        [self.delegate childAdded:snapshot atIndex:index];
+        if ([self.delegate respondsToSelector:@selector(array:didAddObject:atIndex:)]) {
```
ianbarber: :+1:

FirebaseUI/Database/Implementation/FirebaseArray.m
```diff
@@ -20,113 +20,142 @@
...
         NSUInteger index = [self indexForKey:snapshot.key];
 
         [self.snapshots removeObjectAtIndex:index];
 
-        [self.delegate childRemoved:snapshot atIndex:index];
+        if ([self.delegate respondsToSelector:@selector(array:didRemoveObject:atIndex:)]) {
+          [self.delegate array:self didRemoveObject:snapshot atIndex:index];
+        }
       }
       withCancelBlock:^(NSError *error) {
```
ianbarber: General question: should we be removing handles here as well from _handles?

FirebaseUI/Database/Implementation/FirebaseCollectionViewDataSource.m
```diff
@@ -234,38 +237,36 @@ - (instancetype)initWithQuery:(FIRDatabaseQuery *)query
...
 }
 
-- (void)childRemoved:(id)obj atIndex:(NSUInteger)index {
+- (void)array:(FirebaseArray *)array didRemoveObject:(id)object atIndex:(NSUInteger)index {
   [self.collectionView
       deleteItemsAtIndexPaths:@[ [NSIndexPath indexPathForRow:index inSection:0] ]];
 }
 
-- (void)childMoved:(id)obj fromIndex:(NSUInteger)fromIndex toIndex:(NSUInteger)toIndex {
+- (void)array:(FirebaseArray *)array didMoveObject:(id)object fromIndex:(NSUInteger)fromIndex toIndex:(NSUInteger)toIndex {
```
ianbarber: Split onto multiple lines for length

FirebaseUI/Database/Implementation/FirebaseTableViewDataSource.m
```diff
@@ -236,45 +235,43 @@ - (instancetype)initWithQuery:(FIRDatabaseQuery *)query
...
-- (void)childRemoved:(id)obj atIndex:(NSUInteger)index {
+- (void)array:(FirebaseArray *)array didRemoveObject:(id)object atIndex:(NSUInteger)index {
   [self.tableView beginUpdates];
   [self.tableView deleteRowsAtIndexPaths:@[ [NSIndexPath indexPathForRow:index inSection:0] ]
                         withRowAnimation:UITableViewRowAnimationAutomatic];
   [self.tableView endUpdates];
 }
 
-- (void)childMoved:(id)obj fromIndex:(NSUInteger)fromIndex toIndex:(NSUInteger)toIndex {
+- (void)array:(FirebaseArray *)array didMoveObject:(id)object fromIndex:(NSUInteger)fromIndex toIndex:(NSUInteger)toIndex {
```
ianbarber: line length

FirebaseUI/Auth/AuthProviderUI/Facebook/Source/FIRFacebookAuthUI.h
morganchen12: Behavioral change. Over all I'd like to discuss how we want this sort of stuff to look like with the community. I personally want the login providers to be dumb, non-mutating objects that are basically data sources that don't actually have any behavior of their own. But we should discuss with others, since my perspective is lacking.

FirebaseUI/Auth/AuthProviderUI/Google/Source/FIRGoogleAuthUI.m
morganchen12: The prefix has been deleted because it was defined before but never used.

FirebaseUI/Auth/AuthUI/Source/FIRAuthUI.h
morganchen12: Nope, did that on accident

FirebaseUITests/FirebaseArrayTest.m
```diff
@@ -0,0 +1,451 @@
+// clang-format off
```
ianbarber: Does clang-format get tests wrong?

FirebaseUITests/FirebaseArrayTestUtils.h
```diff
@@ -0,0 +1,60 @@
+//
```
ianbarber: Header comment

FirebaseUITests/FirebaseArrayTestUtils.m
```diff
@@ -0,0 +1,115 @@
+//
```
ianbarber: Header comment

FirebaseUITests/FirebaseCollectionViewDataSourceTest.m
```diff
@@ -0,0 +1,59 @@
...
+  [super setUp];
+  self.collectionView = [[UICollectionView alloc] initWithFrame:CGRectZero
+                                           collectionViewLayout:[[UICollectionViewFlowLayout alloc] init]];
+  [self.collectionView registerClass:[UICollectionViewCell class]
+          forCellWithReuseIdentifier:kTestReuseIdentifier];
+  self.observable = [[FUITestObservable alloc] init];
+  self.dataSource = [[FirebaseCollectionViewDataSource alloc] initWithRef:(FIRDatabaseReference *)self.observable
+                                                      cellReuseIdentifier:kTestReuseIdentifier
+                                                                          // no really nil is ok
+                                                                     view:(UICollectionView *_Nonnull)nil];
```
ianbarber: :D

FirebaseUI/Database/API/FirebaseArray.h
morganchen12: I think it makes sense to keep `copy` since it'll copy on reads. Our style guide doesn't say anything about property annotations, and in general for an open source project that we'll be breaking/changing a lot I think it's healthier to disregard internal style guides and agree on whatever style we'll use with the community.

ianbarber:
General CL looks good - definitely add a small changes/migration list, and any updates to docs. 

One thing that occurs to me is that the fakes you made for the database primitives may be generally useful for developers. It might be worth considering cleaning them up and moving into a separate subproject so that people can use them in their own tests.

FirebaseUI/Database/API/FirebaseArrayDelegate.h
morganchen12: In our source comments we suggest users use FirebaseArray to build on top of, so I think it's worth treating as public. Also it's extremely confusing to tell users that something exposed by the module as public is actually not supposed to be public.

FirebaseUI/Database/Implementation/FirebaseArray.m
morganchen12: No, but it would definitely result in a bug.

FirebaseUI/Database/Implementation/FirebaseArray.m
morganchen12: Another good thing to bring up in style discussion.

FirebaseUI/Database/Implementation/FirebaseArray.m
morganchen12: Not entirely sure. There's definitely an odd leap of faith going on here, where we take a query that potentially throws errors and wrap it in a magical array object that never handles errors except by passing them to a delegate. If the query throws an error, the array object may potentially become useless. I'd need to know more about the kinds of errors we throw to be sure.

FirebaseUITests/FirebaseArrayTest.m
morganchen12: That's just the thing for the license in the header that I've been copy pasting

FirebaseUI/Database/Implementation/FirebaseArray.m
morganchen12: Also the FirebaseArray never cancels its own query, so if that callback were to happen unexpectedly, it'd probably be some awful shared mutable state bug where the query was also being used somewhere else.

Maybe it's worth making queries immutable with mutable subclasses and NSCopying, like the NSURLSessionTask family of objects. But that change would have to happen in upstream Firebase.

FirebaseUI/Database/Implementation/FirebaseArray.m
morganchen12: caught red-handed copy pasting code

morganchen12:
@ianbarber this PR now includes tests for basically all of databaseUI, which imo is good enough to merge and try to get CI up and running. Take a look at this when you can and I'll eventually put together a comprehensive changelist before merging.

Thanks!

FirebaseUI.xcodeproj/project.pbxproj
iulukaya: pod-deintegrate doesn't delete those. You can simply delete those 3 build phases.

morganchen12:
Changelog:

FirebaseUI Changes

- FIRFacebookAuthUI scopes became readonly
- FIRFacebookAuthUI initializers no longer return optionals
- FIRGoogleAuthUI initializers no longer return optionals
- FIRGoogleAuthUI: Added designated initializer which takes an array of scopes along with client ID
- FIRGoogleAuthUI: Updated default scopes to use full G+ profile and email scope URLs
- FIRGoogleAuthUI: Made scopes property readonly
- FIRAuthUI: `authUI` method was renamed to `defaultAuthUI`
- FirebaseArray: `delegate` property is now nullable
- FirebaseArray: `initWithQuery:` parameter must be nonnull
- FirebaseArrayDelegate: methods renamed to take callee as first parameter, following platform conventions
    - `childAdded:atIndex:` renamed to `array:didAddObject:atIndex:`
    - `childChanged:atIndex:` renamed to `array:didChangeObject:atIndex:`
    - `childRemoved:atIndex:` renamed to `array:didRemoveObject:atIndex:`
    - `childMoved:fromIndex:toIndex:` renamed to `array:didMoveObject:fromIndex:toIndex:`
    - `cancelledWithError:` renamed to `array:queryCancelledWithError:`
- FirebaseCollectionViewDataSource: added `null_resettable` to `cellClass` and `modelClass`
- FirebaseCollectionViewDataSource: made `populateCell` readonly
- FirebaseTableViewDataSource: added `null_resettable` to `modelClass`
- FirebaseTableViewDataSource: made `populateCell` readonly
- FirebaseDataSource: made `initWithArray:` designated initializer


This only includes breaking and non-breaking API changes, not bug fixes.

FirebaseUI/Database/API/FirebaseArray.h
ianbarber: I don't believe copy copies on read, only write. 

Regarding the style, but in the absence of a known community opinion we should prefer existing style guides. If we diverge from that, we should note that we are, and document our own style variations. 

Regardless of what we choose, we should be consistent - so either change this one to match the order in https://github.com/firebase/FirebaseUI-iOS/pull/108/files#diff-13a906499860a93cf34277f908dfb0b7R37 or vice versa.

FirebaseUI/Database/Implementation/FirebaseArray.m
ianbarber: Nice idea! May be worth writing up an internal API change proposal there if you're interested - Sam has done it before and can walk you through.

FirebaseUI/Auth/AuthProviderUI/Facebook/Source/FIRFacebookAuthUI.h
ianbarber: +1 to that view. It may be worth flagging, just in case people have been constructing and then setting scopes.

FirebaseUI/Auth/AuthProviderUI/Google/Source/FIRGoogleAuthUI.m
```diff
@@ -60,17 +65,22 @@ @implementation FIRGoogleAuthUI {
...
+  return [self initWithClientID:clientID
+                         scopes:@[kGoogleUserInfoEmailScope, kGoogleUserInfoProfileScope]];
+}
+
+- (instancetype)initWithClientID:(NSString *)clientID scopes:(NSArray *)scopes {
   self = [super init];
   if (self) {
     _clientID = [clientID copy];
-    _scopes = @[ @"email", @"profile" ];
+    _scopes = scopes;
```
ianbarber: This is a bit of an example of the concern from before - you're assigning to scopes, and skipping the "copy" property attribute. I think the right thing to do is to make the public property nonatomic, readonly, and in the anonymous category in the .m file override it to nonatomic, copy, readwrite.

FirebaseUI/Database/API/FirebaseArray.h
morganchen12: The implementation manually copies on read, since the backing array is mutable. I guess I could just delete copy since it doesn't do anything here.

FirebaseUI/Database/Implementation/FirebaseArray.m
ianbarber: OK. More visible failure is good. Might be worth a really explicit error message.

morganchen12:
Gonna merge for now, and then we can decide on the accessors in initializers style rule later.
As the codebase currently is, all of the auth code uses ivars in initializers and all of database doesn't. :((((((((((

PR #112, opened 2016-08-19
Dynamic Frameworks are only supported on >= iOS 8

kylebshr:
Dynamic frameworks were introduced with iOS 8, and I believe cocoapods `use_frameworks!` will also only work on iOS 8 or greater.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ pod 'FirebaseUI/Google', '~> 0.4'
 If you're including FirebaseUI in a Swift project, make sure you also have:
 
 ```
-platform :ios, '7.0'
+platform :ios, '8.0'
 use_frameworks!
 ```
 


kylebshr:
I signed it!

morganchen12:
FirebaseUI does support iOS 7 iirc, but that's because there are some giant hacks in place (if you look into the repo, there's no framework targets even though `use_frameworks!` works). This minimum version bump and abandoning the giant hacks is something I'd like to do in the very near future.

kylebshr:
@morganchen12 that's very interesting. I also wasn't trying to say the project isn't compatible with iOS 7, just that you normally can't install with `use_frameworks!` on iOS 7. But you're saying the project actually *will* work on iOS 7 with `use_frameworks!`? If so, I'll close the PR :)

morganchen12:
Ah, I see what you mean. `use_frameworks!` should fail on iOS 7 regardless of our hacks if we're dynamically linking, since iOS 7 just doesn't support dylibs. I'd have to take a closer look and see what's going on, but I'm hoping to override all this stuff and just bump our minimum version to iOS 8 anyway.

morganchen12:
Yeah I lied, FirebaseUI doesn't support iOS 7 (nor should it). I'll merge this ahead of time to not mislead people.

Thanks @kylebshr!

yorgosg:
So, FirebaseUI no longer supports iOS 7? Is this only for Swift?

PR #115, opened 2016-09-06
Change static library targets to dynamic framework targets

morganchen12:
Previously our assets weren't bundled with the static libs we were building, meaning in order to release anything we'd have to bundle them manually.

This PR doesn't solve everything--the dylibs need to be `lipo`d to run on both simulator and device, and then they need to be un`lipo`d (and codesigned) when embedding them in the product since the [App Store rejects apps that link against fat dylibs](http://www.openradar.me/19209161).


morganchen12:
Looks like we won't need to strip slices out because of CocoaPods magic. This PR now mostly works now: the `build.swift` script generates the correct folder structure with assets and everything. The only remaining thing is I broke CI somehow :(

FirebaseUI.podspec
```diff
@@ -1,11 +1,12 @@
...
   s.name         = 'FirebaseUI'
-  s.version      = '0.5.0'
+  s.version      = '0.5.1'
   s.summary      = 'UI binding libraries for Firebase.'
   s.homepage     = 'https://github.com/firebase/FirebaseUI-iOS'
   s.license      = { :type => 'Apache 2.0' }
   s.author       = 'Firebase'
-  s.source       = { :http => 'https://dl.google.com/dl/firebase/firebaseui/ios/0_5_0/FirebaseUIFrameworks.zip' }
+  # s.source       = { :http => 'https://dl.google.com/dl/firebase/firebaseui/ios/0_5_0/FirebaseUIFrameworks.zip' }
+  s.source       = { :http => 'http://localhost:7777/FirebaseUIFrameworks.zip' }
```
mcdonamp: Is this an intentional check in?

mcdonamp: Also, you should talk to @yramanchuk about his changes here for local development.

morganchen12: oops

README.md
```diff
@@ -1,25 +1,23 @@
...
+FirebaseUI is an open-source library for iOS that allows you to quickly connect common UI elements to the [Firebase](https://firebase.google.com) database for data storage, allowing views to be updated in realtime as they change, and providing simple interfaces for common tasks like displaying lists or collections of items.
```
mcdonamp: Sweet mother of UTM campaigns. Let's keep a `utm_source=FirebaseUI-iOS` at the minimum... (but yeah, who added the other ones?)

FirebaseUI/Auth/README.md
```diff
@@ -1,217 +0,0 @@
-# FirebaseUI for iOS — Auth
```
dakeshi: Would you want to remove docs(auth/database) or relocate them?

morganchen12:
Tests pass! Quick, merge while no one is looking

yramanchuk:
1. don't forget to update 2nd podspec (dev one) after merge
2. I'm not sure, but usually we are not removing dependent lib versions (FBSDKLoginKit', '~> 4.0')
I had issues, when 3rd party was updated before our release and already tested version got broken

PR #117, opened 2016-09-07
fix development pods configuration

yramanchuk:
Fix configuration of development cocoa pods for objective-c sample application

### DIFF
--- a/FirebaseUI_dev.podspec
+++ b/FirebaseUI_dev.podspec
@@ -1,0 +1,56 @@
+Pod::Spec.new do |s|
+  s.name         = 'FirebaseUI'
+  s.version      = '0.5.0'
+  s.summary      = 'UI binding libraries for Firebase.'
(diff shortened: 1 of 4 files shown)


FirebaseUI_dev.podspec
```diff
@@ -0,0 +1,56 @@
+Pod::Spec.new do |s|
+  s.name         = 'FirebaseUI'
+  s.version      = '0.5.0'
+  s.summary      = 'UI binding libraries for Firebase.'
+  s.homepage     = 'https://github.com/firebase/FirebaseUI-iOS'
+  s.license      = { :type => 'Apache 2.0' }
+  s.author       = 'Firebase'
+  s.source       = {
+    :git => 'https://github.com/firebase/FirebaseUI-iOS.git',
+    :tag => "v#{s.version}"}
```
morganchen12: Why not use a local git for development instead of a remote? That way you don't need to push your changes to see them. Also, always fetching from the last tagged release means your new commits won't appear when you `pod install`.

PR #118, opened 2016-09-08
Fix uidemo error and refactoring codes about UIStoryboard

dakeshi:
- remove FIRAuthUI from import in AppDelegate
- Add UIStoryboard Extension


morganchen12:
Thanks @dakeshi!

Issue #146, opened 2016-09-21
Extend SDWebImage to accept a FIRStorageReference


mcdonamp:
Add support for Firebase Storage image loading by extending [SDWebImage](https://github.com/rs/SDWebImage).

Sample code:
```
// Load from imageRef into UIImageView imageView
[imageView sd_setImageWithStorageReference:imageRef
                          placeholderImage:[UIImage imageNamed:@"placeholder.png"]
                                 completed:^(UIImage *image, NSError *error, SDImageCacheType cacheType, NSURL *imageURL) {
                                // ... completion code here ...
                           }];
```

morganchen12:
Done in #150.

Archivus:
FIRStorage doesn't seem to support `imageView.sd_setImage(with: <#T##URL?#>, placeholderImage: <#T##UIImage?#>, options: <#T##SDWebImageOptions#>)`

morganchen12:
No. Use one of the methods in [`UIImageView+FirebaseStorage.h`](https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseStorageUI/UIImageView%2BFirebaseStorage.h) instead.

Archivus:
@morganchen12 None of the offered methods seem to support `SDWebImageOptions`. I was looking to download progressive JPEGs using the` SDWebImageOptions.progressive` option.

morganchen12:
I see. You can open a new issue for this, but adding support for SDWebImageOptions is very involved and likely won't happen for a while.

admingj:
Did SDWebImageOptions ever get supported?

morganchen12:
No, FirebaseUI Storage doesn't support any loading via SDWebImage yet unfortunately.

admingj:
Okay thank you. Is there any way to recreate what Firebase is doing with the StorageReference? Aka can we turn it into a url somehow? Or does FirebaseUI create some sort of encrypted url path that it uses to download from Cloud Storage (that isn't replicable with a plain string Url)?

Only way I know of right now is of course doing the `getSignedUrl` method but that takes another trip to the server (or using Cloud Functions with a Service Account)

https://stackoverflow.com/questions/50706401/firebase-function-node-js-to-return-download-url-of-image-via-filename-from

morganchen12:
The easiest way to do this would be to just generate a download URL using the Storage SDK and then pass it into SDWebImage, at which point you'd get everything for free.

PR #150, opened 2016-09-22
Storage

morganchen12:
Addresses #146.


morganchen12:
@mcdonamp please review

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,147 @@
...
+@property (nonatomic, readonly, nullable) id<FUIDownloadTask> sd_currentDownload;
+
+/**
+ * Sets the image view's image to an image downloaded from the Firebase Storage reference.
+ *
+ * @param storageRef  A Firebase Storage reference containing an image.
+ * @return Returns a FIRStorageDownloadTask if a download was created (i.e. image
+ *   could not be found in cache).
+ */
+- (nullable id<FUIDownloadTask>)sd_setImageWithStorageReference:(id<FUIStorageReference>)storageRef;
```
mcdonamp: document size restriction (5MB limit). Also consider increasing to 10MB (Parse had 10MB file limits for instance). Let's aim as high as we can go without allowing OOMs

mcdonamp: I also suggest having a global "setDefaultMaxImageSize" method so it's less opaque that we're using a default size/the dev can easily set it.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+                                     completion:completion];
+}
+
+- (id<FUIDownloadTask>)sd_setImageWithStorageReference:(id<FUIStorageReference>)storageRef
+                                           maxImageSize:(UInt64)size
+                                       placeholderImage:(nullable UIImage *)placeholder
+                                             completion:(void (^)(UIImage *,
+                                                                  NSError *,
+                                                                  SDImageCacheType,
+                                                                  id<FUIStorageReference>))completion{
```
mcdonamp: id<FUIStorageReference>))completion {

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,147 @@
...
+ * @param size        The maximum size of the downloaded image. If the downloaded image
+ *   exceeds this size, an error will be raised in the completion block.
+ * @param placeholder An image to display while the download is in progress.
+ * @param cache       An image cache to check for images before downloading.
+ * @param completion  A closure to handle events when the image finishes downloading.
+ *   The closure is not guaranteed to be invoked on the main thread.
+ * @return Returns a FIRStorageDownloadTask if a download was created (i.e. image
+ *   could not be found in cache).
+ */
+- (nullable id<FUIDownloadTask>)sd_setImageWithStorageReference:(id<FUIStorageReference>)storageRef
```
mcdonamp: NS_DESIGNATED_INITIALIZER

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  // Set placeholder image
+  self.image = placeholder;
+
+  // Query cache for image before trying to download
+  NSString *key = storageRef.fullPath;
+  UIImage *cached = nil;
+
+  cached = [cache imageFromMemoryCacheForKey:key];
```
mcdonamp: Consider breaking out into separate method

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+  cached = [cache imageFromDiskCacheForKey:key];
+  if (cached != nil) {
+    self.image = cached;
+    if (completion != nil) {
+      completion(cached, nil, SDImageCacheTypeDisk, storageRef);
+    }
+    return nil;
+  }
+
+  // If nothing was found in cache, download the image from Firebase Storage
```
mcdonamp: Consider breaking out into separate method

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+                                                  completion:^(NSData * _Nullable data,
+                                                               NSError * _Nullable error) {
+    if (data != nil) {
+      dispatch_async(dispatch_get_main_queue(), ^{
+        self.sd_currentDownload = nil;
+        UIImage *image = [UIImage sd_imageWithData:data];
+        self.image = image;
+
+        // Cache downloaded image
+        [cache storeImage:image forKey:storageRef.fullPath];
```
mcdonamp: Consider cache invalidation for different download tokens, I'd make sure the full path also has the ?token=<download_token> otherwise the images may be different.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+
+  // Query cache for image before trying to download
+  NSString *key = storageRef.fullPath;
+  UIImage *cached = nil;
+
+  cached = [cache imageFromMemoryCacheForKey:key];
+  if (cached != nil) {
+    self.image = cached;
+    if (completion != nil) {
+      completion(cached, nil, SDImageCacheTypeMemory, storageRef);
```
mcdonamp: where is the completion callback raised? Is that always on the main thread (specified by SDWebImage?)?

FirebaseStorageUI/UIImageView+FirebaseStorage.m
mcdonamp: Also I feel like this may be doable on a background thread, pending how long it takes to save this to the cache. Technically all that needs to happen on the main thread is the UIImage set.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+        if (completion != nil) {
+          completion(image, nil, SDImageCacheTypeNone, storageRef);
+        }
+      });
+    } else {
+      dispatch_async(dispatch_get_main_queue(), ^{
+        self.sd_currentDownload = nil;
+      });
+      if (completion != nil) {
+        completion(nil, error, SDImageCacheTypeNone, storageRef);
```
mcdonamp: Same issue here--curious if SDWebImage raises the callback on the main thread for you or if this should go in the above.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,146 @@
...
+      }
+    }
+  }];
+  self.sd_currentDownload = download;
+  return download;
+}
+
+#pragma mark - Accessors
+
+- (void)setSd_currentDownload:(id<FUIDownloadTask>)currentDownload {
```
mcdonamp: sd_setCurrentDownload

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,147 @@
...
+@protocol FUIImageCache <NSObject>
+@required
+
+- (nullable UIImage *)imageFromMemoryCacheForKey:(NSString *)key;
+- (nullable UIImage *)imageFromDiskCacheForKey:(NSString *)key;
+- (void)storeImage:(UIImage *)image forKey:(NSString *)key;
+
+@end
+
+@protocol FUIDownloadTask <NSObject>
```
mcdonamp: What's the purpose of these protocols? If it's for testing, just use OCMock and forget about implementing all of these.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,174 @@
...
+//
+
+@import XCTest;
+@import FirebaseStorageUI;
+
+#import "UIImageView+FirebaseStorage.h"
+
+@class MockDownloadTask, MockStorageReference, MockCache;
+
+@interface MockDownloadTask : NSObject <FUIDownloadTask>
```
mcdonamp: Yeah, we should use OCMock here since it'll make this code *way* more readable.

FirebaseStorageUI/UIImageView+FirebaseStorage.h
morganchen12: These aren't initializer methods, though a designator of some sort would be cool

FirebaseUI.podspec
```diff
@@ -1,11 +1,12 @@
...
   s.name         = 'FirebaseUI'
-  s.version      = '0.5.6-rc3'
+  s.version      = '0.5.6-rc4'
   s.summary      = 'UI binding libraries for Firebase.'
   s.homepage     = 'https://github.com/firebase/FirebaseUI-iOS'
   s.license      = { :type => 'Apache 2.0', :file => 'FirebaseUIFrameworks/LICENSE' }
   s.author       = 'Firebase'
-  s.source       = { :http => 'https://github.com/firebase/FirebaseUI-iOS/releases/download/v0.5.6-rc3/FirebaseUIFrameworks.zip' }
+  # s.source       = { :http => 'https://github.com/firebase/FirebaseUI-iOS/releases/download/v0.5.6-rc3/FirebaseUIFrameworks.zip' }
+  s.source = { :http => 'http://localhost:7777/FirebaseUIFrameworks.zip' }
```
morganchen12: Yeah, these podspec edits aren't going to be part of the final PR, I just left them around on accident.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
morganchen12: The caching is up to SDImageCache, which does disk operations on a background queue.

FirebaseStorageUI/UIImageView+FirebaseStorage.h
mcdonamp: I'm fine if we leave them, but they should move to the test class rather than being in the public interface.

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Sets the image view's image to an image downloaded from the Firebase Storage reference.
+ *
+ * @param storageRef  A Firebase Storage reference containing an image.
+ * @param placeholder An image to display while the download is in progress.
+ * @return Returns a FIRStorageDownloadTask if a download was created (i.e. image
+ *   could not be found in cache).
+ */
+- (nullable FIRStorageDownloadTask *)sd_setImageWithStorageReference:(FIRStorageReference *)storageRef
+                                                placeholderImage:(UIImage *)placeholder;
```
mcdonamp: fix colon alignment

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,142 @@
...
+          completion(nil, error, SDImageCacheTypeNone, storageRef);
+        }
+      }
+    });
+  }];
+  self.sd_currentDownloadTask = download;
+  return download;
+}
+
+#pragma mark - Accessors
```
mcdonamp: Might want to leave a comment here about why we're using associated objects, namely: http://stackoverflow.com/questions/12252725/defining-a-property-in-ios-class-extension

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,188 @@
...
+}
+
+@end
+
+
+@interface MockStorageReference : NSObject <FUIStorageReference>
+@property (nonatomic, readwrite) NSString *fullPath;
+@end
+
+@implementation MockStorageReference
```
mcdonamp: I'll do this one for you, since it's worth comparison to OCMock:
```
// Always use
id mockStorageReference = OCMClassMock([FIRStorageReference class]);

// Test failure case
NSError *error = [NSError errorWithDomain:@"FUITestDownloadErrorDomain" code:1 userInfo:nil];
void (^errorCallback)(NSData *, NSError *) = [OCMArg invokeBlockWithArgs:[NSNull null], error, nil];
OCMStub([mockStorageReference dataWithMaxSize:[OCMArg any] callback:errorCallback])

// Test success case
void (^successCallback)(NSData *, NSError *) = [OCMArg invokeBlockWithArgs:[NSData new], [NSNull null], nil];
OCMStub([mockStorageReference dataWithMaxSize:[OCMArg any] callback:errorCallback])
```

Podfile
```diff
@@ -66,6 +75,11 @@ target 'Database' do
   pod 'Firebase/Database'
 end
 
+target 'Storage' do
```
mcdonamp: What's the difference between "Storage" and "FirebaseStorageUI"?

FirebaseStorageUI/UIImageView+FirebaseStorage.m
mcdonamp: Ok, cool, thanks for checking. I'm ok with using the path locally, though I have some concerns about staleness in the cache if the image is re-uploaded. We should verify this though.

Podfile
```diff
@@ -1,16 +1,16 @@
...
 
-target 'FirebaseUI' do
-  use_frameworks!
+use_frameworks!
 
-  pod 'Firebase'
-  pod 'Firebase/Database'
-  pod 'Firebase/Auth'
+target 'FirebaseUI' do
+  # pod 'Firebase'
```
mcdonamp: We can just delete these if they're being included in the individual ones. Curious what FirebaseUI does (it seems like it's just a container now).

FirebaseStorageUI/UIImageView+FirebaseStorage.m
morganchen12: It's pretty standard at this point and SDWebImage does this too, so I'm not really worried about it

FirebaseStorageUI/UIImageView+FirebaseStorage.h
mcdonamp: Ah, yeah, duh--it just reminds me of the same pattern...

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: I don't doubt that mocking makes the code less verbose, but it increases coupling/complexity for the sake of less boilerplate. Here you seem to be arguing that you'd rather have complexity than verbosity.

Podfile
morganchen12: deleted

FirebaseStorageUI/UIImageView+FirebaseStorage.m
mcdonamp: Sure, but it's it always easier to add a comment now for future folks than assume knowledge (even of things that are "common")

Podfile
morganchen12: FirebaseStorageUI is a dynamic framework target, Storage is a static lib. Currently our build thing builds the dynamic framework and then swaps the executable with a lipo'd static lib so that way we don't have to bundle the framework manually.

As soon as Firebase ships as a dylib, we can just delete these static lib targets.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
morganchen12: The cache is exposed to the consumer, so they can always manually invalidate as necessary.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
morganchen12: If the code is ever needed elsewhere I will. Even with this stuff here the whole method is like 20 lines

FirebaseStorageUI/UIImageView+FirebaseStorage.m
morganchen12: Here we're always dispatching to the main thread, I'd imagine SDWebImage does as well because we're handling UIKit classes

FirebaseStorageUI/UIImageView+FirebaseStorage.m
mcdonamp: Sure, that's just a bad experience to tell a developer that they should do something manually when we could do it for them (we know when they upload to the same path). I wonder if it's worth wrapping `putFile` and `putData` with `putFileAndInvalidateCache...` or something similar.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
mcdonamp: I don't believe that--if I call this on a background thread, the completion will be raised on that thread--we're never kicking it back to the main thread. I'd go ahead and either wrap this (and every callback) in `dispatch_async(dispatch_get_main_queue(), ^{});` or (even better, though probably less useful, so we don't have to implement it now), let the developer specify a callback queue and use that one everywhere.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
mcdonamp: We're basically just reimplementing mocking by adding a bunch of new protocols and classes that junk up the public interface, and adding additional code that we have to maintain. It also raises the barrier to adding new features to mock, since the developer will have to add new protocols/mock classes vs just stubbing them with a mocking framework.

I think it introduces some complexity (developers have to know OCMock), but that is offset by the standardization of the toolset (they should learn it, or they'll know it) as well as the power (partial mocks are a great example of this that we'd have a hard time implementing without rewriting a lot of code).

I'd much rather take on a well known and maintained dependency that developers will be familiar with than try to rebuild out own that solves temporary needs, especially when the barrier to adding the framework is relatively low. Long term, I think we'll be much happier using OCMock to mock other pieces of our system than trying to reimplement everything ourselves.

Podfile
mcdonamp: Well then... Any chance we could take some time (not now, but in the future, ideally before 1.0) to rename them to explain the difference (maybe `FirebaseUIStorageStatic` and `FirebaseUIStorageDynamic`)?

morganchen12: The differences are highlighted in Xcode, which is where they're modified anyway. If we rename the dylib target, we have to rename the framework while copying it (since the generated framework is the same as the target name).

They share code, so having these duplicate targets doesn't increase the code maintenance burden. It's just ugly.

morganchen12:
Should be good for another round of review

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,112 @@
...
+@interface UIImageView (FirebaseStorage)
+
+/**
+ * The current download task, if the image view is downloading an image.
+ */
+@property (nonatomic, readonly, nullable) FIRStorageDownloadTask *sd_currentDownloadTask;
+
+/**
+ * Sets the image view's image to an image downloaded from the Firebase Storage reference.
+ *
```
mcdonamp: Document max size requirement here

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,112 @@
...
+ * @param storageRef  A Firebase Storage reference containing an image.
+ * @return Returns a FIRStorageDownloadTask if a download was created (i.e. image
+ *   could not be found in cache).
+ */
+- (nullable FIRStorageDownloadTask *)sd_setImageWithStorageReference:(FIRStorageReference *)storageRef;
+
+/**
+ * Sets the image view's image to an image downloaded from the Firebase Storage reference.
+ *
```
mcdonamp: Document max size requirement here

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,112 @@
...
+ * @param placeholder An image to display while the download is in progress.
+ * @return Returns a FIRStorageDownloadTask if a download was created (i.e. image
+ *   could not be found in cache).
+ */
+- (nullable FIRStorageDownloadTask *)sd_setImageWithStorageReference:(FIRStorageReference *)storageRef
+                                                    placeholderImage:(UIImage *)placeholder;
+
+/**
+ * Sets the image view's image to an image downloaded from the Firebase Storage reference.
+ *
```
mcdonamp: Document max size requirement here

FirebaseStorageUI/UIImageView+FirebaseStorage.m
```diff
@@ -0,0 +1,142 @@
...
+}
+
+- (FIRStorageDownloadTask *)sd_setImageWithStorageReference:(FIRStorageReference *)storageRef
+                                           placeholderImage:(UIImage *)placeholder
+                                                 completion:(void (^)(UIImage *_Nullable,
+                                                                      NSError *_Nullable,
+                                                                      SDImageCacheType,
+                                                                      FIRStorageReference *))completion {
+  return [self sd_setImageWithStorageReference:storageRef
+                                  maxImageSize:5e6 // 5 megabytes
```
mcdonamp: Make 10MB (some of the Apple system images are around this size). 20MB may even be more reasonable.

FirebaseStorageUI/UIImageView+FirebaseStorage.m
mcdonamp: Let's solve cache invalidation in the next review, since I think Android does this as well.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,125 @@
...
+  self.cache = OCMClassMock([SDImageCache class]);
+  self.imageView = [[UIImageView alloc] initWithFrame:CGRectMake(0, 0, 100, 100)];
+}
+
+- (void)tearDown {
+  [super tearDown];
+}
+
+- (void)testItCreatesADownloadTaskIfCacheIsEmpty {
+  FIRStorageReference *ref = OCMClassMock([FIRStorageReference class]);
```
mcdonamp: I think that they generally prefer `id ref = OCMClassMock...` since technically it's actually not of type `FIRStorageReference`

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,125 @@
...
+  FIRStorageDownloadTask *download = [self.imageView sd_setImageWithStorageReference:(FIRStorageReference *)ref
+                                                                        maxImageSize:512
+                                                                    placeholderImage:nil
+                                                                               cache:self.cache
+                                                                          completion:^(UIImage *image,
+                                                                                       NSError *error,
+                                                                                       SDImageCacheType cacheType,
+                                                                                       FIRStorageReference * storageRef) {
+    XCTAssert(self.imageView.image == image, @"expected download to populate image");
+    XCTAssert(error == nil, @"expected successful download to not produce an error");
```
mcdonamp: XCTAssertNotNil(error, ...)

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,125 @@
...
+    .andReturn(OCMClassMock([FIRStorageDownloadTask class]));
+  FIRStorageDownloadTask *download = [self.imageView sd_setImageWithStorageReference:(FIRStorageReference *)ref
+                                                                        maxImageSize:512
+                                                                    placeholderImage:nil
+                                                                               cache:self.cache
+                                                                          completion:^(UIImage *image,
+                                                                                       NSError *error,
+                                                                                       SDImageCacheType cacheType,
+                                                                                       FIRStorageReference * storageRef) {
+    XCTAssert(self.imageView.image == image, @"expected download to populate image");
```
mcdonamp: ACTAssertObjectEqual(self.imageView.image, image, ...)

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,125 @@
...
+  OCMStub([ref fullPath]).andReturn(@"path/to/image.png");
+  [self.imageView sd_setImageWithStorageReference:(FIRStorageReference *)ref
+                                     maxImageSize:512
+                                 placeholderImage:nil
+                                            cache:self.cache
+                                       completion:^(UIImage *image,
+                                                    NSError *error,
+                                                    SDImageCacheType cacheType,
+                                                    FIRStorageReference *storageRef) {
+    XCTAssert(image == nil, @"expected failed download to not return an image");
```
mcdonamp: XCTAssertNil(image, ...)

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,125 @@
...
+                                                                                       SDImageCacheType cacheType,
+                                                                                       FIRStorageReference * storageRef) {
+    XCTAssert(self.imageView.image == image, @"expected download to populate image");
+    XCTAssert(error == nil, @"expected successful download to not produce an error");
+  }];
+  XCTAssert(download != nil, @"expected image view with empty cache to attempt a download");
+}
+
+- (void)testItDoesNotCreateADownloadIfImageIsCached {
+  FIRStorageReference *ref = OCMClassMock([FIRStorageReference class]);
```
mcdonamp: Consider pulling the mock ref up to the top and constructing it in setUp, then resetting it in tearDown

FirebaseUI.podspec
```diff
@@ -5,7 +5,6 @@ Pod::Spec.new do |s|
   s.homepage     = 'https://github.com/firebase/FirebaseUI-iOS'
   s.license      = { :type => 'Apache 2.0', :file => 'FirebaseUIFrameworks/LICENSE' }
   s.author       = 'Firebase'
-  s.source       = { :http => 'https://github.com/firebase/FirebaseUI-iOS/releases/download/v0.5.6-rc3/FirebaseUIFrameworks.zip' }
```
mcdonamp: We need something there, right?

Podfile
mcdonamp: Got it, I guess I'm ok with that.

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,112 @@
...
+ * @param storageRef  A Firebase Storage reference containing an image.
+ * @param placeholder An image to display while the download is in progress.
+ * @param completion  A closure to handle events when the image finishes downloading.
+ *   The closure is not guaranteed to be invoked on the main thread.
```
morganchen12: Devs can by immediately dispatching inside of the block, since they control the block param that's passed in. Currently this method does dispatch the completion to the main queue, I just wanted the documentation to be flexible toward changes in the future.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: I want this assertion to be explicit about pointer equality

FirebaseUI.podspec
morganchen12: oops

FirebaseStorageUITests/FUIImageViewCategoryTests.m
mcdonamp: XCTAssertEqual(...)

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: fixed

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: fixed

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: fixed

FirebaseStorageUI/UIImageView+FirebaseStorage.m
morganchen12: downloading 20MB images on mobile data should be a crime

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Sets the maximum image download size, in bytes.
+ * @param size The new maximum image download size.
+ */
++ (void)sd_setDefaultMaxImageSize:(UInt64)size;
+
+/**
+ * The current download task, if the image view is downloading an image.
+ * Must be invoked on the main queue.
```
mcdonamp: Add @throws about not being involved on the main queue.

FirebaseStorageUI/UIImageView+FirebaseStorage.h
```diff
@@ -0,0 +1,127 @@
...
+ */
+- (nullable FIRStorageDownloadTask *)sd_setImageWithStorageReference:(FIRStorageReference *)storageRef;
+
+/**
+ * Sets the image view's image to an image downloaded from the Firebase Storage reference.
+ * Must be invoked on the main queue.
+ *
+ * @param storageRef  A Firebase Storage reference containing an image.
+ * @param placeholder An image to display while the download is in progress.
+ * @return Returns a FIRStorageDownloadTask if a download was created (i.e. image
```
mcdonamp: add @throws w/behavior about not invoking on the main queue

FirebaseStorageUI/UIImageView+FirebaseStorage.h
mcdonamp: Whoops, too high. Only on the other ones.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
mcdonamp: Bump

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,118 @@
...
+  self.cache = [[SDImageCache alloc] initWithNamespace:@"FirebaseStorageUITests"];
+  [self.cache storeImage:image forKey:self.ref.fullPath];
+  FIRStorageDownloadTask *download = [self.imageView sd_setImageWithStorageReference:self.ref
+                                                                        maxImageSize:4096
+                                                                    placeholderImage:nil
+                                                                               cache:self.cache
+                                                                          completion:nil];
+  XCTAssertNil(download, @"expected image view to not create new download when fetching cached image");
+  XCTAssert(self.imageView.image == image, @"expected image view to use cached image");
+  [self.cache cleanDisk];
```
mcdonamp: Consider moving this to `tearDown` so you don't forget/so you can remove boilerplate

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,118 @@
...
+  UIImage *image = [[UIImage alloc] init];
+  self.cache = [[SDImageCache alloc] initWithNamespace:@"FirebaseStorageUITests"];
+  [self.cache storeImage:image forKey:self.ref.fullPath];
+  FIRStorageDownloadTask *download = [self.imageView sd_setImageWithStorageReference:self.ref
+                                                                        maxImageSize:4096
+                                                                    placeholderImage:nil
+                                                                               cache:self.cache
+                                                                          completion:nil];
+  XCTAssertNil(download, @"expected image view to not create new download when fetching cached image");
+  XCTAssert(self.imageView.image == image, @"expected image view to use cached image");
```
mcdonamp: XCTAssertEqual

FirebaseStorageUITests/FUIImageViewCategoryTests.m
```diff
@@ -0,0 +1,118 @@
...
+  [self.cache cleanDisk];
+  [self.cache clearMemory];
+}
+
+- (void)testItRaisesAnErrorIfDownloadingFails {
+  [self.imageView sd_setImageWithStorageReference:self.ref
```
mcdonamp: Do you really not have to do the expectation thing for async tasks?

```
- (void) test {
    XCTestExpectation *expectation = [self expectationWithDescription:@"testFoo"];
    [Foo fooWithCompletion:^(Bar *bar){
      [expectation fulfill];
    }];

    [self waitForExpectationsWithTimeout:10 handler:^(NSError *error) {
      if (error) {
        NSLog(@"Error: %@", error);
      }
    }];
}
```

FirebaseUI.podspec
```diff
@@ -4,8 +4,8 @@ Pod::Spec.new do |s|
   s.summary      = 'UI binding libraries for Firebase.'
   s.homepage     = 'https://github.com/firebase/FirebaseUI-iOS'
   s.license      = { :type => 'Apache 2.0', :file => 'FirebaseUIFrameworks/LICENSE' }
-  s.author       = 'Firebase'
   s.source       = { :http => 'https://github.com/firebase/FirebaseUI-iOS/releases/download/v0.5.6-rc3/FirebaseUIFrameworks.zip' }
```
mcdonamp: Let's go for 0.6.0 for parity with Android. Not sure if this gets fixed up in the build system.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
mcdonamp: BTW, happy if we don't (it was messy), just surprised that we don't (or maybe we do and the tests are flaky?)

FirebaseStorageUI/UIImageView+FirebaseStorage.h
morganchen12: It doesn't throw any errors, though, it's just undefined behavior. Like how [the UIView methods don't throw](https://developer.apple.com/reference/uikit/uiview).

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: It's only used in this one test, though; if it does appear in multiple tests in the future I'll move it there.

FirebaseStorageUITests/FUIImageViewCategoryTests.m
morganchen12: ¯\\\_(ツ)_/¯

FirebaseUI.podspec
morganchen12: Release bumping is gonna happen separately from this PR.

mcdonamp: Cool, thanks for clarifying.

Issue #129, opened 2016-09-13
App crashes with NSRangeException after inserting record being observed by FirebaseTableViewDataSource


chadpav:
### Environment
  * Using FirebaseUI/Database subspec only 
  * Swift 2.3 / Xcode 8 GM
  * Targeting and deploying to iOS version 9.3.5
  * Firebase SDK version: 3.5.2
  * FirebaseUI version: 0.5.x
  * CocoaPods Version: 1.0.1
  
### The problem:
App crashes after inserting a record into database using FirebaseTableViewDataSource. Can not reproduce on 0.4.0, reliably reproduces on >=0.5.0 (although not every time).

#### Steps to reproduce:
  1. Create FirebaseTableViewDataSource to back a UITableView. (I'm using a queryOrderedByPriority)
  2. Insert a record or maybe up to 15 into the database that will result in Inserting a record into the tableview datasource.
  3. If needed Insert / Delete until it crashes.  Sometimes it's every time an other times it's maybe a dozen inserts in.
  
#### Observed Results:

  * App crashes often but not every time. The Insert does succeed so this is happening when the ChildAdded event is coming back to Insert the new record into the internal Array. I believe the code is expecting to find the PrevChild in the array but for some reason the array doesn't contain the item anymore. This explains the `index 9223372036854775808` and the bounds [0..0] don't make any sense because there should have already been 1 record in that array. I just crashed the app again and the array bounds were 1 off (0...3) when I had 5 records in the database. Note, it took me about 10 tries to crash the app this time.

```
*** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[__NSArrayM insertObject:atIndex:]: index 9223372036854775808 beyond bounds [0 .. 0]'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000105740d85 __exceptionPreprocess + 165
	1   libobjc.A.dylib                     0x00000001051b4deb objc_exception_throw + 48
	2   CoreFoundation                      0x0000000105601cc5 -[__NSArrayM insertObject:atIndex:] + 901
	3   Kitchen Wall                        0x0000000101c7c372 __30-[FirebaseArray initListeners]_block_invoke.185 + 257
	4   Kitchen Wall                        0x0000000101c5402f __43-[FChildEventRegistration fireEvent:queue:]_block_invoke64 + 88
	5   libdispatch.dylib                   0x0000000106391d9d _dispatch_call_block_and_release + 12
[...]

PR #162, opened 2016-10-07
Allow elements to be moved to start of FirebaseArray (potential fix for #129)

ramblex:
When using FirebaseTableViewDataSource I was seeing something like the following when moving an object to the start of a list.

`Assertions: failed: caught "NSRangeException", "*** -[__NSArrayM insertObject:atIndex:]: index 9223372036854775807 beyond bounds [0 .. 8]"`

It looks like it was caused by the `FIRDataEventTypeChildMoved` being sent a `previousChildKey` that doesn't exist in the array (presumably ""). This is similar to #129 and is potentially a fix for that.

### DIFF
--- a/FirebaseDatabaseUI/FirebaseArray.m
+++ b/FirebaseDatabaseUI/FirebaseArray.m
@@ -124,9 +124,11 @@
         [self.snapshots removeObjectAtIndex:fromIndex];
 
         NSUInteger toIndex = 0;
-        NSUInteger prevIndex = [self indexForKey:previousChildKey];
-        if (prevIndex != NSNotFound) {
-          toIndex = prevIndex + 1;
+        if (previousChildKey != nil) {
+            NSUInteger prevIndex = [self indexForKey:previousChildKey];
+            if (prevIndex != NSNotFound) {
+                toIndex = prevIndex + 1;
+            }
         }
         [self.snapshots insertObject:snapshot atIndex:toIndex];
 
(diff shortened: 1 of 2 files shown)


FirebaseDatabaseUI/FirebaseArray.m
```diff
@@ -123,7 +123,11 @@ - (void)initListeners {
         NSUInteger fromIndex = [self indexForKey:snapshot.key];
         [self.snapshots removeObjectAtIndex:fromIndex];
 
-        NSUInteger toIndex = [self indexForKey:previousChildKey] + 1;
+        NSUInteger toIndex = 0;
+        NSUInteger prevIndex = [self indexForKey:previousChildKey];
```
morganchen12: We should be handling the case where the previous child key is `nil` as well here

FirebaseDatabaseUITests/FirebaseArrayTest.m
```diff
@@ -378,4 +378,36 @@ - (void)testFirebaseArrayCanMoveElement {
...
+    self.arrayDelegate.didMoveObject = ^(FirebaseArray *array, id object, NSUInteger from, NSUInteger to) {
+        // Xcode complains about retain cycles if an XCTAssert is placed in here.
+        delegateWasCalled = YES;
+        expectedParametersWereCorrect = (array == self.firebaseArray &&
+                                         object == self.snap &&
+                                         from == 8 && to == 0);
+    };
+    
+    // Move 8 to the start
+    [self.observable sendEvent:FIRDataEventTypeChildMoved withObject:self.snap previousKey:@"" error:nil];
```
morganchen12: Can you add a case like this where previous key is `nil` as well? Other than that this change looks good.

ramblex:
I've added in the changes. I'm not sure exactly why the CI build failed.. there's quite a few `Canceling tests due to timeout in Checking test manager availability` messages.

PR #172, opened 2016-10-13
Custom signin

yramanchuk:
implement custom sign-in controller

### DIFF
--- a/FirebaseAuthUI/FIRPasswordSignInViewController.h
+++ b/FirebaseAuthUI/FIRPasswordSignInViewController.h
@@ -24,31 +24,59 @@ NS_ASSUME_NONNULL_BEGIN
     @brief The view controller that asks for user's password.
  */
 @interface FIRPasswordSignInViewController : FIRAuthUIBaseViewController
+{
+  @protected
+  /** @var _email
+   @brief The @c email address of the user from the previous screen.
+   */
+  NSString *_email;
+
+}
 
 /** @fn initWithNibName:bundle:authUI:
-    @brief Please use @c initWithAuthUI:email:.
+    @brief Please use @c initWithNibName:bundle:authUI:email:.
  */
 - (instancetype)initWithNibName:(nullable NSString *)nibNameOrNil
                          bundle:(nullable NSBundle *)nibBundleOrNil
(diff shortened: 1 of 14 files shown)


FirebaseAuthUI/FIRPasswordSignInViewController.h
```diff
@@ -24,31 +24,59 @@ NS_ASSUME_NONNULL_BEGIN
     @brief The view controller that asks for user's password.
  */
 @interface FIRPasswordSignInViewController : FIRAuthUIBaseViewController
+{
+  @protected
+  /** @var _email
+   @brief The @c email address of the user from the previous screen.
+   */
+  NSString *_email;
```
morganchen12: Why not use a property?

samples/objc/FirebaseUI-demo-objc/Samples/Auth/FIRAuthViewController.m
```diff
@@ -81,11 +81,13 @@ - (void)viewWillAppear:(BOOL)animated {
...
 
+  [self.navigationController setToolbarHidden:NO animated:YES];
 }
 
--(void)viewDidDisappear:(BOOL)animated {
-  [super viewDidDisappear:animated];
+-(void)viewWillDisappear:(BOOL)animated {
+  [super viewWillDisappear:animated];
   [self.auth removeAuthStateDidChangeListener:self.authStateDidChangeHandle];
+  [self.navigationController setToolbarHidden:YES animated:YES];
```
morganchen12: You should probably pass the `animated` flag through to `setToolbarHidden:animated:`.

```objc
[self.navigationController setToolbarHidden:YES animated:animated];
```

That way your code doesn't create animations if they're not going to be visible anyway.

FirebaseAuthUI/FIRPasswordSignInViewController.h
yramanchuk: I don't want to expose this property to any other classes; only parent classes should have access to it.

samples/objc/FirebaseUI-demo-objc/Samples/Auth/FIRAuthViewController.m
yramanchuk: fixed, thanks

PR #185, opened 2016-10-31
Brazilian portuguese translation of Firebase UI

danielpassos:
pt-BR translation port from https://github.com/firebase/FirebaseUI-Android/pull/384

PS: Not 100% sure if I should commit the `FirebaseUI.xcodeproj/project.pbxproj` file


morganchen12:
Thanks for doing this! I'm not qualified to fully review it, but you absolutely should commit the `pbxproj` file, since if you don't Xcode won't know that those files exist. The reason the diff is so big is because we usually run [`pod deintegrate`](https://rubygems.org/gems/cocoapods-deintegrate) before committing.

danielpassos:
@morganchen12 That was the exactly that tip I was looking for. How include the (translated) files without include all pod things ;)

Updated. Thanks for the tip!

morganchen12:
FYI @drishi94, @yramanchuk

morganchen12:
Closing this as localizations have been added by the auth team and we don't have a process for accepting external localizations. Sorry!

Issue #183, opened 2016-10-26
Add support for FUIArray and data source transformations


morganchen12:
This is something that's been requested frequently (#93, #169). Two things I definitely want to add are `map` and `sort`. Not sure if we should add `filter`, since it may encourage people to download much more data than their app ever intends on displaying.

This most likely won't make it into the FirebaseUI 1.0.0 release.

morganchen12:
Preliminary work on this can be found [here](https://github.com/firebase/FirebaseUI-iOS/compare/master...morganchen12:transforms).

Goals:
- FirebaseArray and friends abstracted behind `FIRCollection` protocol, so consumers don't have to care about the details of their collections if they don't want to. Users will also be able to define their own `FIRCollection`s.
- Remove query observation from initialization of `FIRCollection`s. This way initializers don't have side effects.
- Change data sources' designated initializers to take `FIRCollection`s. This way a data source can be initialized with a normal, sorted, or mapped array without concern over the details of the array. Decouples data sources from their storage dependencies. The current initializers can be kept around, but this is still a breaking change since people who subclass the data sources will have to override a different initializer. Data sources are now responsible for telling their collections to observe their respective queries.

mcdonamp:
Let's make sure we have a corresponding bug filed for Android as well.

muirandy:
+1 
This sounds exactly what I'm looking for. 

I'm wanting to display a tableView, however I'm unable to get exactly what I need form the firebase query, as its either not sorted or not filtered - I can't have them both on the query itself (a well known issue/feature of Firebase!).

I want to be able to create my own implementation of FUIArray, which I can sort and filter as needed. At the moment, I can't do this without also writing my own DataSource.

_> "Two things I definitely want to add are map and sort. Not sure if we should add filter, since it may encourage people to download much more data than their app ever intends on displaying."_

Please do implement filter as well as sort! I understand the concern regarding performance and the download size, however, at least in my case this isn't going to be an issue - I'm dealing with relatively small amounts of data.

morganchen12:
@muirandy The PR is basically ready to be merged, we're just waiting on Android since we want to maintain feature parity between the two codebases. Even though filter is left unimplemented, `FIRCollection` is publicly available and you can conform your own filtered collection to it.

morganchen12:
#186 has been merged. Release coming soon.

morganchen12:
v2.0.0 has been released.

cgossain:
[...]

PR #186, opened 2016-10-31
Transforms

morganchen12:
Add support for defining transformations (i.e. sort) on the data from FirebaseDatabase before it's presented in a table/collection view. Still missing a few things. Addresses #183, depends on ~~#184~~. Also allows for consumers to create their own collections.

- [x] ~~Doc comments~~
- [x] ~~DRY code in collections~~
- [x] ~~Tests~~

@mcdonamp please review


morganchen12:
cc @puf, who may have opinions on this

FirebaseDatabaseUI/FUICollection.h
```diff
@@ -18,57 +18,109 @@
...
  */
-@protocol FUIArrayDelegate<NSObject>
+@protocol FUICollectionDelegate<NSObject>
 
 @optional
 
 /**
+ * Called before any other events are sent.
+ */
+- (void)arrayDidBeginUpdates:(id<FUICollection>)collection;
```
puf: I intentionally did not add a "begin updates" event, but only a "we're done with all updates". What do you think you would catch in this method that you would catch with a simple "is this the first update?" check?

morganchen12: This is just to better mirror the UIKit aggregation methods of [`beginUpdates`](https://developer.apple.com/reference/uikit/uitableview/1614908-beginupdates) and [`endUpdates`](https://developer.apple.com/reference/uikit/uitableview/1614890-endupdates).

morganchen12: I suspect it's ok to drop the `beginUpdates` event if the platform doesn't call for it. :)

PR #195, opened 2016-11-04
Fix FBSDKLoginKit not being visible by default to dyld

morganchen12:
Addresses #194


morganchen12:
Gonna close this, as I wasn't able to reproduce the error in #194 in either of our samples or our unit tests target (targets that depend on FirebaseFacebookAuthUI).

PR #221, opened 2016-12-12
Update Readme.md to reflect new Facebook SDK docs

astromme:
It appears that the URL schemes are now step 4 in Facebook's documentation, not "following step 5".


morganchen12:
Thanks!

Issue #128, opened 2016-09-13
Runtime error in call to authViewController(): 'Could not load NIB in bundle: 'NSBundle ... (loaded)' with name 'FIRAuthPickerViewController''


davidsantiago:
- Objective C or Swift: Swift
- iOS version: 9.3.5
- Firebase SDK version: 3.4.0
- FirebaseUI version: 0.5.4
- CocoaPods Version: 1.1.0-rc1
- Xcode version: 8.0 GM
### Step 3: Describe the problem:

Sorry to be back here so soon after you were so quick on the turnaround with the other issue I posted earlier. 

Now that 0.5.4 will build correctly, I get an exception about the bundle not being able to open the NIB file "FIRAuthPickerViewController" at runtime when FIRAuthUI.authViewController() is called. 
#### Steps to reproduce:
1. Code below fails at runtime. This happens in my own project, but I see that it also happens in the sample code in this repo, so I'll reference that. Start the uidemo project, select the Auth demo, hit Start, and it'll crash immediately.
#### Observed Results:

  The code compiles and runs, but it throws an exception at runtime during the call to authUI!.authViewController(). 

Here is the barf at runtime: 

2016-09-12 17:08:00.576 uidemo[6783:3230924] **\* Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Could not load NIB in bundle: 'NSBundle </var/containers/Bundle/Application/5A387BD1-3913-41CB-A5E1-88A7306AC0E9/uidemo.app> (loaded)' with name 'FIRAuthPickerViewController''
**\* First throw call stack:
(0x18333adb0 0x18299ff80 0x18333acf8 0x1889733a4 0x188815230 0x1885da118 0x1885a8ea4 0x18849c8ec 0x18849c844 0x100093b40 0x1000967d4 0x10009bdc8 0x10002426c 0x10002455c 0x1884d4be8 0x1884d4b64 0x1884bc870 0x1884d4454 0x1884d4084 0x1884ccc20 0x18849d04c 0x18849b628 0x1832f109c 0x1832f0b30 0x1832ee830 0x183218c50 0x184b00088 0x188506088 0x100032430 0x182db68b8)
libc++abi.dylib: terminating with uncaught exception of type NSException
#### Expected Results:

The function to successfully return the ViewController. 
#### Relevant Code:

[...]

PR #225, opened 2016-12-21
use hard-coded string instead of class name for nib

morganchen12:
@yramanchuk FYI

Fixes an issue where subclassing would cause `[self class]` to return a string that wasn't a valid nib. Addresses #128.


morganchen12:
I think it's okay to hard-code a nib name here because it's not the designated initializer, so devs who want to use their own nibs can override this method in their subclasses. Also I'd prefer hard-coding the nib name here because it provides the least surprising behavior when subclassing.

In either case the codebase continues to work, but IMO without the hard-coded nib name third-party devs must do more work to create the expected behavior when subclassing without a custom nib.

PR #339, opened 2017-09-22
Adding temporary workaround to call FirebaseAuth completion block in …

yramanchuk:
Workaround to fix https://github.com/firebase/firebase-ios-sdk/issues/304

### DIFF
--- a/FirebasePhoneAuthUI/FUIPhoneEntryViewController.m
+++ b/FirebasePhoneAuthUI/FUIPhoneEntryViewController.m
@@ -168,7 +168,7 @@ static NSString *const kNextButtonAccessibilityID = @"NextButtonAccessibilityID"
   [provider verifyPhoneNumber:phoneNumberWithCountryCode
                    UIDelegate:self
                    completion:^(NSString *_Nullable verificationID, NSError *_Nullable error) {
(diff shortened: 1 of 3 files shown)


FirebasePhoneAuthUI/FUIPhoneEntryViewController.m
```diff
@@ -168,28 +168,36 @@ - (void)onNext:(NSString *)phoneNumber {
...
-      
-      FUIPhoneAuth *delegate = [self.authUI providerWithID:FIRPhoneAuthProviderID];
-      [delegate callbackWithCredential:nil error:error result:nil];
-      return;
+      [self pushViewController:controller];
+    };
+    if ([NSThread isMainThread]) {
+      completionBlock();
+    } else {
+      dispatch_sync(dispatch_get_main_queue(), completionBlock);
```
XiangtianDai: I missed this earlier... I think we'd better to do dispatch_async instead of dispatch_sync, otherwise if the delegate calls FIRAuth methods there is a risk of deadlock.

morganchen12: Fixed in 1eb511c2430546295fc5b6a03786511f71932a0a

Issue #380, opened 2017-12-21
Twitter updated their framework's name 


SirArkimedes:
* Objective C or Swift: Swift
  * iOS version: 11.2
  * Firebase SDK version: 4.8.0
  * FirebaseUI version: 4.5.0
  * CocoaPods Version: 1.3.1
  
### Problem:

Twitter updated their framework name and ```FirebaseTwitterAuthUI``` uses the old name. This is a breaking change if ```TwitterKit``` & ```TwitterCore``` are updated and FirebaseUI is included in the project.

#### Steps to reproduce:

  1. Clone this repo
  2. pod install
  3. Start a build of ```FirebaseTwitterAuthUI```
  4. Notice that ```TwitterKit.h``` is not found
  
#### Observed Results:

  * Get a "```TwitterKit/TwitterKit.h``` is not found" error.
  
#### Expected Results:

  * FirebaseUI works with latest ```TwitterKit``` & ```TwitterCore```.

I am willing to submit a PR to fix this issue:
#381

morganchen12:
Nice catch. PRs are always welcome, if not I'll fix this in the new year since I'm on vacation soon.

For the time being, people who run into this issue can work around it by pinning the Twitter SDK to an older version.

morganchen12:
This is fixed in the latest release.

PR #381, opened 2017-12-22
Support changed Twitter.h to TWTRTwitter.h

SirArkimedes:
Fixes issue #380. If ```FirebaseUI``` is included in a project that runs ```pod  update```, an error no longer occurs from FirebaseUI with this change.

```TwitterCore``` is necessary to be imported on ```FUITwitterAuth.m``` because ```TWTRLogInErrorCodeCancelled``` is in ```TwitterCore```.


morganchen12:
I'm not sure this fixes the `pod update` issue though, since all we did was change header imports in implementation files and FirebaseUI is distributed as a binary.

SirArkimedes:
@morganchen12 Hmm. I pulled from my fork with Cocoapods and it was fine from my change. Are you seeing it still being an issue when it is in binary form? I'm not sure how to include a binary version with Cocoapods.

morganchen12:
I'll take a closer look then, don't worry about it. 👍

PR #382, opened 2017-12-26
Add textContentType for UITextField

xuaninbox:
- add textContentType for UITextField to support iOS 11 QuickType keyboard


xuaninbox:
I signed it!

PR #383, opened 2017-12-28
Add textContentType for UITextField

xuaninbox:
- add textContentType for UITextField to support iOS 11 QuickType keyboard


FirebasePhoneAuthUI/FUIPhoneEntryViewController.m
```diff
@@ -254,6 +254,7 @@ - (UITableViewCell *)tableView:(UITableView *)tableView
     _phoneNumberField.autocapitalizationType = UITextAutocapitalizationTypeNone;
     _phoneNumberField.returnKeyType = UIReturnKeyNext;
     _phoneNumberField.keyboardType = UIKeyboardTypeNumberPad;
+    _phoneNumberField.textContentType = UITextContentTypeTelephoneNumber;
```
protocol86: `UITextContentTypeTelephoneNumber` seems to be only available in iOS 10+.  Should we add a check for iOS 10 here?

Issue #139, opened 2016-09-19
Anonymous account automatic conversion to permanent account


alfongj:
Description: if this feature were implemented, FirebaseUI would have the capability of linking account credentials to an already signed-in anonymous account.

As a sub-requirement, FirebaseUI should also expose an API to handle the case where the credentials the user tries to link to correspond to an already existing account, so the developer can handle the merge case manually.

hijamoya:
+1 need this!

codemodouk:
+1 also need this!

SebastianBO:
+1

yorgosg:
So, If this were implemented, I guess it would by default have the same behavior as now, unless the developer uses the exposed API, to handle it? Otherwise, it would break stuff in previous implementations, right?

ibru:
+1

I've actually began implementing the feature myself.
But I'm having problems use my fork as Pod and test my implementation. I was not even able to build forked sources. If anyone would help me with this we might be able to put it together.

I'm solving this by adding another delegate method which will be able to grab credentials and decide if FirebaseUI controller should proceed with singing in user or not. This should handle also the sub-requirement mentioned above.

My work is here: https://github.com/ibru/FirebaseUI-iOS/tree/anonymous-link

morganchen12:
@ibru take a look at how Firebase is integrated into the ObjC sample.

Firebase is shipped as a static framework, so CocoaPods doesn't ordinarily let us use it as a dependency of a framework. We hack around this limitation by just shipping binaries, but it means testing changes locally becomes really painful.

SUPERCILEX:
@morganchen12 Any chance you could work on this? My PR on the Android side (https://github.com/firebase/FirebaseUI-Android/pull/309) has been waiting for months (Sep 2016). Thanks!

morganchen12:
Yeah, I'll try to take a look soon.

SUPERCILEX:
That would be amazing, thanks!

yorgosg:
@morganchen12 I guess there would be a way to not use the automatic conversion, right? I have done a different implementation to handle this and my code might break if this is done automatically in a future version.

morganchen12:
Yeah, automatic conversion would be an opt-in feature.

ibru:
Heads up that I've began working on this change. My current progress is in this branch https://github.com/ibru/FirebaseUI-iOS/commits/anonymous-link

[...]

PR #401, opened 2018-02-01
Anonymous user upgrade federated

protocol86:
Anonymous user upgrade federated accounts and phone auth. 

Swift sample will be updated in a  follow up PR.

Addresses #139

### DIFF
--- a/FirebaseAuthUI/FUIAuth.h
+++ b/FirebaseAuthUI/FUIAuth.h
@@ -190,10 +190,11 @@ __attribute__((deprecated("Instead use authUI:didSignInWithAuthDataResult:error:
  */
 @property(nonatomic, copy, nullable) NSURL *TOSURL;
 
-/** @property autoUpgradeAnonymousUsers
+/** @property shouldAutoUpgradeAnonymousUsers
     @brief Whether to enable auto upgrading of anonymous accounts, defaults to NO.
  */
-@property(nonatomic, assign, getter=isAutoUpgradeAnonymousUsers) BOOL autoUpgradeAnonymousUsers;
+@property(nonatomic, assign, getter=shouldAutoUpgradeAnonymousUsers) BOOL
+    shouldAutoUpgradeAnonymousUsers;
 
 /** @property delegate
     @brief A delegate that receives callbacks or provides custom UI for @c FUIAuth.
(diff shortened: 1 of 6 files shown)


FirebaseAuthUI/FUIAuth.h
```diff
@@ -190,6 +190,11 @@ __attribute__((deprecated("Instead use authUI:didSignInWithAuthDataResult:error:
  */
 @property(nonatomic, copy, nullable) NSURL *TOSURL;
 
+/** @property autoUpgradeAnonymousUsers
+    @brief Whether to enable auto upgrading of anonymous accounts, defaults to NO.
+ */
+@property(nonatomic, assign, getter=isAutoUpgradeAnonymousUsers) BOOL autoUpgradeAnonymousUsers;
```
morganchen12: nit: rename this to `shouldAutoUpgradeAnonymousUsers` (for both getter and setter)

samples/objc/FirebaseUI-demo-objc/Samples/Auth/FUIAuthViewController.m
```diff
@@ -279,21 +310,9 @@ - (NSString *)getAllIdTokens {
 
 - (void)signOut {
   NSError *error;
-  FIRUser *currentUser = self.authUI.auth.currentUser;
-  // If the user is anonymous, delete the user to avoid dangling anonymous users.
-  if (currentUser.isAnonymous) {
-    [currentUser deleteWithCompletion:^(NSError * _Nullable error) {
```
morganchen12: Do we no longer need to delete anonymous users?

protocol86: We do, I just moved the code to the tableview cell. The reason is that I wanted to use the sign in button to trigger a sign-in when the current user was anonymous. Essentially I am separating anonymous sign in/out with "normal" sign in/out.

morganchen12:
#403 is now merged, feel free to merge this PR as soon as travis tests pass.

Thanks!

FirebaseAuthUI/FUIAuth.h
protocol86: Thanks, done.

PR #407, opened 2018-02-14
Adds anonymous upgrade for google account email conflict

protocol86:
Adds logic to link a federated account with an existing google account if they share the same email during anonynous user upgrade.

### DIFF
--- a/FirebaseAuthUI/FUIAuth.m
+++ b/FirebaseAuthUI/FUIAuth.m
@@ -213,7 +213,6 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
             NSError *mergeError = [NSError errorWithDomain:FUIAuthErrorDomain
                                                       code:FUIAuthErrorCodeMergeConflict
                                                   userInfo:userInfo];
-            result(nil, mergeError);
             completeSignInBlock(authResult, mergeError);
           } else if (error.code == FIRAuthErrorCodeEmailAlreadyInUse) {
             if ([providerUI respondsToSelector:@selector(email)]) {
@@ -223,7 +222,6 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
                              completion:^(FIRAuthDataResult *_Nullable authResult,
                                           NSError *_Nullable error) {
                 if (error) {
-                  result(nil, error);
                   completeSignInBlock(nil, error);
                   return;
                 }
@@ -232,7 +230,6 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
                                                                          *_Nullable authResult,
                                                                      NSError *_Nullable error) {
                   if (error) {
-                    result(nil, error);
                     completeSignInBlock(nil, error);
                     return;
                   }
@@ -243,7 +240,6 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
                   NSError *mergeError = [NSError errorWithDomain:FUIAuthErrorDomain
                                                              code:FUIAuthErrorCodeMergeConflict
                                                          userInfo:userInfo];
-                  result(nil, mergeError);
                   completeSignInBlock(authResult, mergeError);
                 }];
               }];
@@ -325,7 +321,7 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
 
 - (nullable NSString *)federatedAuthProviderFromProviders:(NSArray <NSString *> *) providers {
   NSSet *providerSet =
-      [[NSSet alloc] initWithArray:@[ FIRFacebookAuthProviderID, FIRGoogleAuthProviderID ]];
+      [NSSet setWithArray:@[ FIRFacebookAuthProviderID, FIRGoogleAuthProviderID ]];
   for (NSString *provider in providers) {
     if ( [providerSet containsObject:provider]) {
       return provider;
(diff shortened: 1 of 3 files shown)


FirebaseAuthUI/FUIAuth.m
```diff
@@ -210,11 +210,44 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
...
+          } else if (error.code == FIRAuthErrorCodeEmailAlreadyInUse) {
+            if ([providerUI respondsToSelector:@selector(email)]) {
+              // Link federated providers
+              [self signInWithEmailHint:[providerUI email]
+               presentingViewController:presentingViewController
+                             completion:^(FIRAuthDataResult *_Nullable authResult,
+                                          NSError *_Nullable error) {
+                if (error) {
+                  result(nil, error);
+                  completeSignInBlock(nil, error);
```
morganchen12: The comment above also applies here.

FirebaseAuthUI/FUIAuth.m
```diff
@@ -210,11 +210,44 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
...
+                  return;
+                }
+                [authResult.user linkAndRetrieveDataWithCredential:credential
+                                                        completion:^(FIRAuthDataResult
+                                                                         *_Nullable authResult,
+                                                                     NSError *_Nullable error) {
+                  if (error) {
+                    result(nil, error);
+                    completeSignInBlock(nil, error);
```
morganchen12: See above comment

FirebaseAuthUI/FUIAuth.m
```diff
@@ -210,11 +210,44 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
...
+                  }
+                  FIRAuthCredential *newCredential = credential;
+                  NSDictionary *userInfo = @{
+                    FUIAuthCredentialKey : newCredential,
+                  };
+                  NSError *mergeError = [NSError errorWithDomain:FUIAuthErrorDomain
+                                                             code:FUIAuthErrorCodeMergeConflict
+                                                         userInfo:userInfo];
+                  result(nil, mergeError);
+                  completeSignInBlock(authResult, mergeError);
```
morganchen12: See above comment

FirebaseAuthUI/FUIAuth.m
```diff
@@ -250,6 +283,57 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
...
+                      presentingViewController:presentingViewController
+                                    completion:^(FIRAuthCredential *_Nullable credential,
+                                                 NSError *_Nullable error,
+                                                 FIRAuthResultCallback  _Nullable result) {
+        if (error) {
+          completion(nil, error);
+          return;
+        }
+
+        [auth signInAndRetrieveDataWithCredential:credential completion:completion];
```
morganchen12: Why does this use a different auth instance?

FirebaseAuthUI/FUIAuth.m
```diff
@@ -250,6 +283,57 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
...
+      }];
+    }
+  }];
+}
+
+- (nullable NSString *)federatedAuthProviderFromProviders:(NSArray <NSString *> *) providers {
+  NSSet *providerSet =
+      [[NSSet alloc] initWithArray:@[ FIRFacebookAuthProviderID, FIRGoogleAuthProviderID ]];
```
morganchen12: nit: use `+[NSSet setWithArray:]`

Also, using `NSSet` for a faster contains when searching an array is not necessarily faster than using `containsObject:` on the array directly, since some cost is incurred when creating the set.

FirebaseAuthUI/FUIAuth.m
```diff
@@ -210,11 +210,44 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
...
               FUIAuthCredentialKey : newCredential,
             };
-            mergeError = [NSError errorWithDomain:FUIAuthErrorDomain
-                                             code:FUIAuthErrorCodeMergeConflict
-                                         userInfo:userInfo];
+            NSError *mergeError = [NSError errorWithDomain:FUIAuthErrorDomain
+                                                      code:FUIAuthErrorCodeMergeConflict
+                                                  userInfo:userInfo];
             result(nil, mergeError);
             completeSignInBlock(authResult, mergeError);
```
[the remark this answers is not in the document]
protocol86: that's a good catch, we seemed to have forgotten to remove this call after adding the completeSignInBlock.

FirebaseAuthUI/FUIAuth.m
protocol86: done

FirebaseAuthUI/FUIAuth.m
protocol86: done

FirebaseAuthUI/FUIAuth.m
protocol86: done

FirebaseAuthUI/FUIAuth.m
protocol86: Because we need to complete a sign-in of the existing credential before linking; however we do not want to lose the currently signed in user.  A sign-in operation usually replaces the currently signed-in user if successful.

FirebaseAuthUI/FUIAuth.m
protocol86: Agreed, using the class method but leaving the set, just don't like the idea of creating an array for the sole purpose of look up :).

FirebaseAuthUI/FUIAuth.m
morganchen12: Maybe add a comment describing that :)

protocol86: Done.

PR #410, opened 2018-02-23
Adds email/password “email-already-in-use” flow when upgrading to a federated credential

protocol86:
Adds flow to sign into an existing email/password user (on a separate auth instance)  before linking this user to the federated auth credential during a merge conflict while upgrading anonymous user.

### DIFF
--- a/FirebaseAuthUI/FUIAuth.m
+++ b/FirebaseAuthUI/FUIAuth.m
@@ -331,7 +331,9 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
                                                  NSError *_Nullable error,
                                                  FIRAuthResultCallback  _Nullable result) {
           if (error) {
-            completion(nil, error);
+            if (completion) {
+              completion(nil, error);
+            }
             return;
           }
           [tempAuth signInAndRetrieveDataWithCredential:credential completion:completion];
--- a/FirebaseAuthUI/FUIAuth.m
+++ b/FirebaseAuthUI/FUIAuth.m
@@ -313,7 +313,7 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
         // Email password sign-in
         FUIPasswordSignInViewController *controller =
             [[FUIPasswordSignInViewController alloc] initWithAuthUI:authUI email:emailHint];
-        [controller setOnDismissCallback:completion];
+        controller.onDismissCallback = completion;
         [presentingViewController pushViewController:controller];
       } else {
         id<FUIAuthProvider> authProviderUI;
--- a/FirebaseAuthUI/FUIAuth.m
+++ b/FirebaseAuthUI/FUIAuth.m
@@ -28,6 +28,7 @@
 #import "FUIAuthStrings.h"
 #import "FUIGoogleAuth.h"
 #import "FUIEmailEntryViewController.h"
+#import "FUIPasswordSignInViewController_Internal.h"
 #import "FUIPasswordVerificationViewController.h"
 
 /** @var kAppNameCodingKey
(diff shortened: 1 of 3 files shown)


FirebaseAuthUI/FUIPasswordSignInViewController.m
```diff
@@ -134,7 +139,11 @@ - (void)signInWithDefaultValue:(NSString *)email andPassword:(NSString *)passwor
         }
       }
       [self.navigationController dismissViewControllerAnimated:YES completion:^{
-        [self.authUI invokeResultCallbackWithAuthDataResult:authResult error:error];
+        if (self->_onDismissCallback) {
+          self->_onDismissCallback(authResult, error);
```
XiangtianDai: It is usually a good idea to nil-out callback blocks after use to avoid circular references.

FirebaseAuthUI/FUIPasswordSignInViewController_Internal.h
```diff
@@ -0,0 +1,35 @@
...
+
+@interface FUIPasswordSignInViewController ()
+
+/** @fn setOnDismissCallback:
+    @brief Sets an optional custom callback for FUIPasswordSigInViewController during dismissal. If
+        this callback is set the default dismissal routine is not triggered and should be included
+        in this block if necessary.
+    @param callback The custom callback to execute during dismissal of the view controller.
+ */
+- (void)setOnDismissCallback:(FIRAuthDataResultCallback)callback;
```
XiangtianDai: I would just make it a property called `completion`, with initial value being the what we're doing by default. If you do that, make sure you only capture weak self in the block.

Alternatively you can use the delegate pattern but that might be too heavy-handed if you don't expect more methods in the delegate.

FirebaseAuthUI/FUIPasswordSignInViewController.m
```diff
@@ -59,6 +59,11 @@ @implementation FUIPasswordSignInViewController {
       @brief The @c UIButton which handles forgot password action.
    */
   __weak IBOutlet UIButton *_forgotPasswordButton;
+
+  /** @var _onDismissCallback
+      @brief The callback to be executed during view controller dismissal.
+   */
+  FIRAuthDataResultCallback _onDismissCallback;
```
morganchen12: A better name might be `onSuccessfulLoginCallback`, since this block will not be called if the controller is dismissed by exiting the login flow.

FirebaseAuthUI/FUIPasswordSignInViewController.m
protocol86: I went with the property approach outlined in you other comment. Since the block will be reused I leave it up to the developer using a custom block to decide whether or not it should be set to nil.

FirebaseAuthUI/FUIPasswordSignInViewController_Internal.h
protocol86: Done, thanks!

FirebaseAuthUI/FUIAuth.m
```diff
@@ -291,47 +292,60 @@ - (void)signInWithEmailHint:(NSString *)emailHint
...
+          }
         }
+        [authProviderUI signOut];
+        [authProviderUI signInWithDefaultValue:emailHint
+                      presentingViewController:presentingViewController
+                                    completion:^(FIRAuthCredential *_Nullable credential,
+                                                 NSError *_Nullable error,
+                                                 FIRAuthResultCallback  _Nullable result) {
+          if (error) {
+            completion(nil, error);
```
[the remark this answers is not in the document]
protocol86: Done.

FirebaseAuthUI/FUIAuth.m
```diff
@@ -280,7 +281,7 @@ - (void)signInWithProviderUI:(id<FUIAuthProvider>)providerUI
 }
 
 - (void)signInWithEmailHint:(NSString *)emailHint
-   presentingViewController:(UIViewController *)presentingViewController
+   presentingViewController:(FUIAuthBaseViewController *)presentingViewController
```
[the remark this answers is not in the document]
protocol86: `signInWithEmailHint:presentingViewController:completion:` isn't a public API, it's only used for a special case of automatic upgrade behind the scenes. The `pushViewController:` method being used here is a custom method which modifies the way the navigation bar is presented when the view controller is pushed.

PR #414, opened 2018-03-02
Handles email mismatch case

protocol86:
Handles the case of email mismatch when the developer is prompted to sign-in with an existing account.

Does not include localization for new UI messages. These will be added in a separate PR.


FirebaseAuthUI/FUIAuth.m
```diff
@@ -31,6 +31,18 @@
 #import "FUIPasswordSignInViewController_Internal.h"
 #import "FUIPasswordVerificationViewController.h"
 
+/** @typedef EmailHintSignInCallback
+    @brief The type of block invoked when an emailHint sing-in event completes.
```
morganchen12: nit: `sign-in` typo

PR #432, opened 2018-04-11
Handles email mismatch case and fixes bug for federated credential already in use.

protocol86:
Handles email mismatch case and fixes bug for federated credential already in use.

### DIFF
--- a/FirebaseAuthUI/FUIAuth.m
+++ b/FirebaseAuthUI/FUIAuth.m
@@ -235,7 +235,7 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
                           originalError:error
                              completion:^(FIRAuthDataResult *_Nullable authResult,
                                           NSError *_Nullable error,
-                                          FIRAuthCredential *_Nullable credential) {
+                                          FIRAuthCredential *_Nullable existingCredential) {
                 if (error) {
                   completeSignInBlock(nil, error);
                   return;
@@ -325,7 +325,9 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
   [self.auth fetchProvidersForEmail:emailHint completion:^(NSArray<NSString *> *_Nullable providers,
                                                            NSError *_Nullable error) {
     if (error) {
-      completion(nil, error, nil);
+      if (completion) {
+        completion(nil, error, nil);
+      }
       return;
     }
     NSString *existingFederatedProviderID = [self authProviderFromProviders:providers];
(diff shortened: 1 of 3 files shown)


FirebaseAuthUI/FUIAuth.m
```diff
@@ -31,6 +31,18 @@
...
 
+/** @typedef EmailHintSignInCallback
+    @brief The type of block invoked when an emailHint sign-in event completes.
+
+    @param authResult Optionally; Result of sign-in request containing both the user and
+       the additional user info associated with the user.
+    @param error Optionally; the error which occurred - or nil if the request was successful.
+    @param credential Optionally; The credential used to sign-in.
+ */
+typedef void (^EmailHintSignInCallback)(FIRAuthDataResult *_Nullable authResult,
```
morganchen12: Should this be `FUIEmailHintSignInCallback`?

FirebaseAuthUI/FUIAuth.m
```diff
@@ -299,7 +325,7 @@ - (void)signInWithEmailHint:(NSString *)emailHint
   [self.auth fetchProvidersForEmail:emailHint completion:^(NSArray<NSString *> *_Nullable providers,
                                                            NSError *_Nullable error) {
     if (error) {
-      completion(nil, error);
+      completion(nil, error, nil);
```
morganchen12: Is `completion` guaranteed to be nonnull here?

FirebaseAuthUI/FUIAuth.m
```diff
@@ -351,16 +381,49 @@ - (void)signInWithEmailHint:(NSString *)emailHint
...
+              }
+
+              // Handle potential email mismatch.
+              if (![emailHint isEqualToString:authResult.user.email]) {
+                NSString *signedInEmail = authResult.user.email;
+                NSString *title =
+                    [NSString stringWithFormat:@"Continue sign in with %@?", signedInEmail];
+                NSString *message =
+                    [NSString stringWithFormat:@"You originally wanted to sign in with %@",
+                    emailHint];
```
morganchen12: If these strings are user-facing, they should be localized.

FirebaseAuthUI/FUIAuth.m
```diff
@@ -351,16 +381,49 @@ - (void)signInWithEmailHint:(NSString *)emailHint
...
+                [FUIAuthBaseViewController showAlertWithTitle:title
+                                                      message:message
+                                                  actionTitle:@"Continue"
```
morganchen12: This string also requires localization.

FirebaseAuthUI/FUIAuth.m
protocol86: I didn't think we need the prefix here since it's an internal callback.

FirebaseAuthUI/FUIAuth.m
protocol86: The entire flow needs to be localized in a follow up CL.

FirebaseAuthUI/FUIAuth.m
protocol86: Will localize in follow up CL.

FirebaseAuthUI/FUIAuth.m
protocol86: Good catch, fixed.

PR #457, opened 2018-05-24
Email entry title

### LINKED ISSUE #125: Differentiate between logging in and signing up
Hi,

I would like to put in a feature request to differentiate between signing in and signing up.  Right now the buttons always say "Sign In". It would be really nice to pass in a parameter that allowed us to set the language to "Log in" or "Sign up".

This request is based off of user feedback, people have come to our sign up screen and asked where do I Sign Up?

Thank you!
Laura

najdanovicivan:
Changed Email Entry Title fix #125


morganchen12:
Thanks!

PR #479, opened 2018-07-15
Added some space after markdown headers tags

amadeu01:
Added space after `###`, so the hearder could look like markdown header


amadeu01:
I signed it!

morganchen12:
Thanks!

PR #481, opened 2018-07-17
Automatic anonymous user upgrade

protocol86:
This feature allows an anonymous user to automatically upgrade to a non-anonymous user upon sign-in. It handles several edge cases associated with mismatched credentials.

This PR does not contain internationalized strings for mismatch messages. These will be added in a follow up PR.

### DIFF
--- a/FirebaseAuthUI/FUIAuth.h
+++ b/FirebaseAuthUI/FUIAuth.h
@@ -201,6 +201,12 @@ __attribute__((deprecated("Instead use authUI:didSignInWithAuthDataResult:error:
  */
 @property(nonatomic, copy, nullable) NSURL *TOSURL;
 
+/** @property shouldAutoUpgradeAnonymousUsers
+    @brief Whether to enable auto upgrading of anonymous accounts, defaults to NO.
+ */
+@property(nonatomic, assign, getter=shouldAutoUpgradeAnonymousUsers) BOOL
+    shouldAutoUpgradeAnonymousUsers;
+
 /** @property privacyPolicyURL
     @brief The URL of your app's Privacy Policy. If not nil, a privacy policy notice is
         displayed on the initial sign-in screen and potentially the phone number auth and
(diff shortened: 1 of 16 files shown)


Yue-Wang-Google:
The code looks good to me. Possible improvements may include split the two long sign in functions in FUIAuth.m to smaller chunks to improve readability. Doesn't needs to happen in this CL, though.

FirebaseAuthUI/FUIAuth.h
```diff
@@ -201,6 +201,12 @@ __attribute__((deprecated("Instead use authUI:didSignInWithAuthDataResult:error:
  */
 @property(nonatomic, copy, nullable) NSURL *TOSURL;
 
+/** @property shouldAutoUpgradeAnonymousUsers
+    @brief Whether to enable auto upgrading of anonymous accounts, defaults to NO.
+ */
+@property(nonatomic, assign, getter=shouldAutoUpgradeAnonymousUsers) BOOL
```
Yue-Wang-Google: why the getter is needed?

FirebaseAuthUI/FUIAuthBaseViewController.m
```diff
@@ -242,20 +242,61 @@ + (void)showAlertWithTitle:(nullable NSString *)title
...
+             cancelHandler:(FUIAuthAlertActionHandler)cancelHandler {
+  UIAlertController *alertController =
+      [UIAlertController alertControllerWithTitle:title
+                                          message:message
+                                   preferredStyle:UIAlertControllerStyleAlert];
+  UIAlertAction *okAction =
+      [UIAlertAction actionWithTitle:actionTitle
+                               style:UIAlertActionStyleDefault
+                             handler:^(UIAlertAction *_Nonnull action) {
+        actionHandler();
```
Yue-Wang-Google: I understand those handler are with nonnull type. but is it a good practice to check null before calling them? same to other handler calls.

FirebaseAuthUI/FUIAuth.h
protocol86: This is a public API, the developer may want to condition on whether shouldAutoUpgrade is turned on or not.

Yue-Wang-Google: @protocol86    I mean the property name and the getter name are identical (shouldAutoUpgradeAnonymousUsers). why getter=?

FirebaseAuthUI/FUIAuthBaseViewController.m
protocol86: ooh, missed this. Will be addressed in a follow up PR.

radvansky-tomas:
Could we split shouldAutoUpgradeAnonymousUsers for sign in and sign up ? 
I mean I want to upgrade users only during sign up process, and if they already have account its causing merging errors !

PR #498, opened 2018-08-07
Conform TOS/PP view to safe area insets

### LINKED ISSUE #439: Sign In buttons doesn't constrain to Safe Area
I realized that the Sign In buttons are constrained to the superview rather than the Safe area. This causes it to look strange on the iPhone X. Is there a way to fix this? Thanks.

morganchen12:
Fixes #439


morganchen12:
Don't merge, looks like this breaks iOS 8 builds.

Yue-Wang-Google:
Looks good to me. Thanks!

mikezing12:
icon-app-29x29 3x

wittawats:
opportunity

morganchen12:
Closing since the change is smaller than the merge conflict, will re-create this PR when we drop iOS 8.

Issue #484, opened 2018-07-20
Updating to 5.1.2 breaks phone and Google auth


lswang1618:
Objective C or Swift:  Swift
iOS version: 11.4
Firebase SDK version:  FirebaseCore 5.0.5
FirebaseUI version: 5.1.2
CocoaPods Version: 1.5.2
  
### Step 3: Describe the problem:
Logging in using FirebaseUI with Google stalls on loading after selecting email without useful console messages. Tapping on Phone from the auth picker screen results in . Downgrading to 5.0.0 resolves both issues. 

#### Steps to reproduce:

  1. Tap Sign in with Google
  2. Tap Continue
  3. Select account
  4. Observe stall via progress circle that never ends (although login and upgrade does seem to complete as observed in Firebase Console)

  1. Tap Sign in with phone
  2. Observe crash
  
#### Observed Results:
When logging in with Google, no useful errors, just stalled loading. 
When trying to log in with Phone, results in this crash: 
![screen shot 2018-07-20 at 12 55 47 am](https://user-images.githubusercontent.com/2426354/42983949-b94ce8d4-8bb7-11e8-911b-ea8cb44df883.png)
  
#### Expected Results:

  * Successful login with either Google or Phone auth
  
#### Relevant Code:

                let authUI = FUIAuth.defaultAuthUI()
                // You need to adopt a FUIAuthDelegate protocol to receive callback
                authUI?.delegate = self
                
                let providers: [FUIAuthProvider] = [
                    FUIGoogleAuth(),
                    FUIPhoneAuth(authUI:FUIAuth.defaultAuthUI()!),
                    ]
                authUI?.providers = providers
                authUI?.shouldAutoUpgradeAnonymousUsers = true
        
                let authViewController = authUI?.authViewController()
                present(authViewController!, animated: true, completion: nil)

morganchen12:
Can you provide the full stack trace of the crash?

morganchen12:
I'm able to log in successfully using Google auth in the swift sample app in this repository.

lswang1618:
For phone auth: 
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0xda9bf7c08)
    frame #0: 0x00000001835bc910 libobjc.A.dylib`objc_msgSend + 16
    frame #1: 0x0000000184e37ddc Foundation`_NSDescriptionWithStringProxyFunc + 64
[...]

PR #499, opened 2018-08-08
Fix issue where decrementActivity wasn't called on email conflict

morganchen12:
This will fix part of #484.


Yue-Wang-Google:
LGTM

PR #531, opened 2018-09-24
Add information about the sign in response.

amadeu01:
I was not sure about how could be possible to get the information about the attempt of the user sign in. So, I remember that the documentation says to set the auth delegate's to the view controller, such as: `authUI?.delegate = self`.

So, in order to get the response of the attempt, I overwrote the delegate method 
```objective-c
- (void)authUI:(FUIAuth *)authUI
    didSignInWithAuthDataResult:(nullable FIRAuthDataResult *)authDataResult
                          error:(nullable NSError *)error;
```

I just wanted to make it more clear for other developers.


morganchen12:
Thanks @amadeu01!

PR #548, opened 2018-10-18
Improve FirestoreUI performance

morganchen12:
Improves performance for diff update after changing query on `FUIBatchedArray`. The diff generated from the new algorithm ay not be entirely minimal, but should always be correct (and doesn't generate egregious amounts of unneeded updates).


FirebaseFirestoreUI/FUISnapshotArrayDiff.m
```diff
@@ -246,7 +246,7 @@ - (instancetype)initWithInitialArray:(NSArray *)initialArray resultArray:(NSArra
   if (self != nil) {
     _initial = [initialArray copy];
     _result = [resultArray copy];
-    [self buildDiffs];
+    [self buildFastDiffs];
```
mikelehen: Is buildDiffs no longer used?  Should it be removed or should there be a comment explaining if/when we might want to revive it?

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
```diff
@@ -377,6 +377,131 @@ - (instancetype)initWithInitialArray:(NSArray<FIRDocumentSnapshot *> *)initial
   return self;
 }
 
+- (void)buildFastDiffs {
```
mikelehen: Is it worth a comment to explain the high-level approach and trade-offs? (you mentioned that it may not be a minimal diff, but it's presumably faster).

Also, is the complexity still O(m*n) ? If not, can you update the comment on the FUISnapshotArrayDiff interface?

FirebaseFirestoreUI/FUISnapshotArrayDiff.h
```diff
@@ -162,8 +162,8 @@ FUIUnorderedPair *FUIUnorderedPairMake(id left, id right);
  * Creates a diff between two arrays. This operation is relatively expensive;
  * O(n x m) for arrays of length n and m.
  */
-- (instancetype)initWithInitialArray:(NSArray<ObjectType> *)initialArray
-                         resultArray:(NSArray<ObjectType> *)resultArray;
+- (instancetype)initWithInitialArray:(NSArray<FIRDocumentSnapshot *> *)initialArray
+                         resultArray:(NSArray<FIRDocumentSnapshot *> *)resultArray;
 
 /**
  * Creates a diff between two arrays, using the document changes array to speed up
```
mikelehen: Unrelated to this PR, but I peeked at this method and I'm pretty confused about what's going on...  Why can't you use documentChanges basically as-is?  There are a bunch of cases handled in the implementation of this method that I can't make sense of (e.g. handling a FIRDocumentChangeTypeRemoved change where the removed document is in the result set for some reason).

If I'm just missing something, consider updating this comment to explain what's going on.

As a further aside (sorry), it's hard for me to reason about the 9 output arrays from this diff (deletedIndexes, deletedObjects, etc...).  It would be nice if the diff output was basically (or literally?) identical to our FIRQuerySnapshot.documentChanges list (just an array of FIRDocumentChange objects).  If there's some reason that structure doesn't work well for you, I'd be curious to hear about it. If it doesn't work for FirebaseUI, then maybe we need to rethink it.  FWIW- I'd be open to exposing the FIRDocumentChange constructor if it allowed interesting API reuse here.

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
```diff
@@ -377,6 +377,131 @@ - (instancetype)initWithInitialArray:(NSArray<FIRDocumentSnapshot *> *)initial
...
+  // collection. Here's the observations we're making:
+  //
+  // 1. If the document is unchanged and its start and end indexes are the same, ignore the
+  //    document.
+  // 2. If the document is different in the initial and result arrays but the documentID
+  //    is the same, then the document was changed (and should be reloaded).
+  // 3. If the document did not exist in the initial array, this corresponds to an add.
+  // 4. If the document did not exist in the result array, this corresponds to a delete.
+  // 5. If an add and delete occurred at the same index, this corresponds to a reload.
+  // 6. If the document exists in both arrays with differing indexes, this corresponds to a move.
```
mikelehen: So to confirm, if I have:

OLD:
1. docA
2. docB

NEW (I removed docA)
1. docB


Then that'll correspond to 2 diffs?
1. docA removed.
2. docB moved from index 2 to 1.

That seems surprising.  If you have a list of 10 items and delete the first one, you'll get 10 diffs since all the indexes change.  Is that what we need / want?

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
```diff
@@ -377,6 +377,131 @@ - (instancetype)initWithInitialArray:(NSArray<FIRDocumentSnapshot *> *)initial
...
+  // to be unique, so we can't have multiple versions of the same document in a
+  // collection. Here's the observations we're making:
+  //
+  // 1. If the document is unchanged and its start and end indexes are the same, ignore the
+  //    document.
+  // 2. If the document is different in the initial and result arrays but the documentID
+  //    is the same, then the document was changed (and should be reloaded).
+  // 3. If the document did not exist in the initial array, this corresponds to an add.
+  // 4. If the document did not exist in the result array, this corresponds to a delete.
+  // 5. If an add and delete occurred at the same index, this corresponds to a reload.
```
mikelehen: For clarity, would it be possible to use the same terminology here as for the actual output arrays. I am assuming that "a reload" corresponds to a "change" (i.e. changedObjects and changedIndexes)?

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
morganchen12: It's dead code and will be removed.

FirebaseFirestoreUI/FUISnapshotArrayDiff.h
```diff
@@ -162,8 +162,8 @@ FUIUnorderedPair *FUIUnorderedPairMake(id left, id right);
  * Creates a diff between two arrays. This operation is relatively expensive;
  * O(n x m) for arrays of length n and m.
  */
-- (instancetype)initWithInitialArray:(NSArray<ObjectType> *)initialArray
-                         resultArray:(NSArray<ObjectType> *)resultArray;
+- (instancetype)initWithInitialArray:(NSArray<FIRDocumentSnapshot *> *)initialArray
+                         resultArray:(NSArray<FIRDocumentSnapshot *> *)resultArray;
```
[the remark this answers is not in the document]
morganchen12: I think that's reasonable.

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
morganchen12: Yeah. It'd be nice to have a perfectly minimal diff but UIKit doesn't seem to mind these so it works for the purposes we're using the diffs for (updating collection/table views).

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
morganchen12: Will do :)

morganchen12:
>Also, out of curiosity, how slow was the old implementation? Were there noticeable hiccups? How
many items did you need for it to become noticeable? It seems like there's a spectrum of diffing tradeoffs that goes from "full LCS minimal diff" on one end to "just trigger removes for all the old items followed by adds for all the new items" on the other end, and I'm wondering a little bit where we should be aiming.

The old implementation bogged down at around 150 elements. You're right in that the new implementation is no longer a minimal diff (which you can see reflected in the changes to the unit tests). The more complex our diff is, the more expensive the UIKit calls will be since UIKit will have to do more updates to end up with the same result state. This change brings us from quadratic performance `m * n` to linear `m + n`, and scales to the point where UIKit is the bottleneck, and not the diffing logic.

FirebaseFirestoreUI/FUISnapshotArrayDiff.h
morganchen12: There's no particularly good reason the `documentChanges` info isn't better-utilized other than  when I wrote it I wasn't as familiar with the guarantees around `documentChanges`. As for the output arrays, it might make more sense in the context of [how they're used](https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseFirestoreUI/FUIFirestoreCollectionViewDataSource.m#L101-L162).

FirebaseFirestoreUI/FUISnapshotArrayDiff.m
morganchen12: Done.

rob-keepsafe:
Might make sense to just use [`DifferenceKit`](https://github.com/ra1028/DifferenceKit) or the [new iOS 13 api's](https://developer.apple.com/videos/play/wwdc2019/220/) for this instead so it's standardized

This might all be a `won't fix` w/ the new iOS 13 apis anyways, but I was hoping you all had an out-of-the-box solution so we wouldn't have to implement this ourselves 😛

Nelsilee (approved):

TwitRco (approved):

Mogiles (approved):
I'm in

Issue #558, opened 2018-11-05
Translations files out of sync


IchordeDionysos:
* Objective C or Swift: all
  * iOS version: all
  * Firebase SDK version: all
  * FirebaseUI version: 5.2.2
  * CocoaPods Version: all
  
### Step 3: Describe the problem:
The English localizable has different translation key-value pairs for privacy policy and terms of service as other languages:
https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseAuthUI/Strings/en.lproj/FirebaseAuthUI.strings#L71
https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseAuthUI/Strings/de.lproj/FirebaseAuthUI.strings#L71

This ends up resulting in for users confusing texts:

![Image showing legal text "TermsOfServiceMessage"](https://trello-attachments.s3.amazonaws.com/5bd91e159043c2344cd6b4af/5bdfc871f5622c40d0412184/b8da47beb7c5749060828b2ca4dcf14b/photo.jpg)

iishabrawy:
Objective C or Swift: swift
iOS version: all
Firebase SDK version: all
FirebaseUI version: 5.2.2
CocoaPods Version: all

same here 

and need to clarify that the device language in arabic or english translation shows like images attached 

![issue1](https://user-images.githubusercontent.com/6254386/49070512-7f22ac00-f234-11e8-9cea-1e76ceab51e5.jpeg)

![issue](https://user-images.githubusercontent.com/6254386/49070337-19362480-f234-11e8-8a06-81997fb3c870.jpeg)

PR #559, opened 2018-11-05
Update de AuthUI strings to match en version

IchordeDionysos:
This addresses #558 unfotunatly only the German version.

I don't want to translate all the other languages with Google Translate or something like this.

Feel free to edit this Pull Request to address all other languages.


Yue-Wang-Google:
should be solved by #592 so closing this

PR #587, opened 2019-01-10
Fix missing icons issue

morganchen12:


Auth/FirebaseAuthUI/FUIAuthUtils.m
```diff
@@ -22,14 +22,16 @@ @implementation FUIAuthUtils
...
   NSBundle *frameworkBundle = nil;
-  if (bundleName) {
-    NSString *path = [[NSBundle mainBundle] pathForResource:bundleName ofType:@"bundle"];
-    if (path == nil) {
-      // Check framework resources if bundle isn't present in main bundle.
-      path = [[NSBundle mainBundle] pathForResource:bundleName ofType:@"framework"];
-    }
-    frameworkBundle = [NSBundle bundleWithPath:path];
-  } else {
+  if (bundleName == nil) {
```
Yue-Wang-Google: avoid  if (foo == nil). use if (!foo). same below.

renkelvin:
Icon of "Sign in as guest" is still missing.

Issue #609, opened 2019-02-05
FUIArray is slow when query has thousands of snapshots


mastohhh:
* Objective C or Swift: Both
  * iOS version: 12
  * Firebase SDK version: 4.0.1
  * FirebaseUI version: 6.0
  * CocoaPods Version: 1.6.1
  
Steps to reproduce:

  1. Create a FUIArray with thousands of results
  
Observed Results:

<img width="810" alt="capture d ecran 2019-02-05 a 16 30 27" src="https://user-images.githubusercontent.com/1760064/52289457-5c1c2300-296e-11e9-8a85-ee7f567fde2e.png">

It takes 8s on simulator and even more on device.

mastohhh:
I created a pull request that optimize `indexForKey` method.

Here's my results : 

<img width="788" alt="capture d ecran 2019-02-05 a 17 28 27" src="https://user-images.githubusercontent.com/1760064/52290460-63dcc700-2970-11e9-9f46-8e5a7f48ec58.png">

morganchen12:
Closed in #610.

PR #610, opened 2019-02-05
optimize FUIArray’s indexForKey method by using an array of keys

mastohhh:
The method `- (NSUInteger)indexForKey:(NSString *)key;` loops in all snapshots and call `key` method on each snapshot and compare them with an `equalToString:` comparison method.

This is very slow as you can see in that issue : https://github.com/firebase/FirebaseUI-iOS/issues/609

The idea here is to maintain an other array of snapshot's keys. The `indexForKey:` method can now call `indexOfObject:` method on that new array. As it's an array of strings, it's very easy and speed to retrieve that index.


mastohhh:
I signed it !

PR #615, opened 2019-02-08
Added missing import of new auth email pod

ghowen:
The import of the new FirebaseUI/Email pod was missing for the sample code to work properly.


ghowen:
I signed it.

morganchen12:
Thanks! I'll assume googlebot is just being slow today.

PR #619, opened 2019-02-14
FUIEmailAuth.m Crash Fix

wimbledon:
If continueURLString is nil, "[NSURLComponents componentsWithString:continueURLString]" will crash.
Return NO for this un-handleOpenURL case.


wimbledon:
My CLA has been submitted.

PR #643, opened 2019-03-12
Add shortName to the OAuth provider initializer.

Yue-Wang-Google:


OAuth/FirebaseOAuthUI/FUIOAuth.h
```diff
@@ -40,6 +40,7 @@ NS_ASSUME_NONNULL_BEGIN
     @param authUI The @c FUIAuth instance that manages controllers of this provider.
     @param providerID The unique identifier for the provider.
     @param buttonLabelText The text label for the sign in button.
+    @param shortName A short display name for the provider.
```
renkelvin: for -> of?

Yue-Wang-Google: The name is pretty arbitrary. so I guess for is more appropriate?

Issue #652, opened 2019-03-26
Spelling mistake in email-link authentication alert box


tndttan:
### Step 2: Describe your environment

  * Objective C or Swift: Swift
  * iOS version: 12.1
  * Firebase SDK version: 5.14.0
  * FirebaseUI version: 6.1.1
  * CocoaPods Version: 1.5.3
  
### Step 3: Describe the problem:

Spelling mistake in 'Sign-in email Sent' alert box, where 'instrucitons' should be 'instructions'.

#### Steps to reproduce:

  1. Choose email-link authentication
  2. Provide an email and tap 'Next'

Yue-Wang-Google:
I'll just take it

PR #653, opened 2019-03-26
fix typo in email link sign in

Yue-Wang-Google:


Yue-Wang-Google:
fixes https://github.com/firebase/FirebaseUI-iOS/issues/652

PR #674, opened 2019-05-06
Fix missing runtime libraries error

morganchen12:


morganchen12:
Closing in favor of #688

Issue #684, opened 2019-05-25
No visible @interface for 'FBSDKLoginManager' declares the selector 'logInWithReadPermissions'


jonathanfoster:
I just updated my pods and ran into a compiler error in FirebaseUI/Facebook. The error is:

```
 No visible @interface for 'FBSDKLoginManager' declares the selector 'logInWithReadPermissions:fromViewController:handler:'
```

It occurs at https://github.com/firebase/FirebaseUI-iOS/blob/master/FacebookAuth/FirebaseFacebookAuthUI/FUIFacebookAuth.m#L139.

Has FirebaseUI 7.0.0 been officially released? As I was researching the issue, I found the latest FirebaseUI release is 6.2.1, but there's a `v7.0.0-prerelease` branch and the Podspec has been updated to 7.0.0 (https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseUI.podspec#L3).

Here's the relevant output from `pod install`:
```
Using FBSDKCoreKit (5.0.1)
Using FBSDKLoginKit (5.0.1)
Using Firebase (6.1.0)
Using FirebaseABTesting (3.0.0)
Using FirebaseAnalytics (6.0.1)
Using FirebaseAnalyticsInterop (1.2.0)
Using FirebaseAuth (6.1.0)
Using FirebaseAuthInterop (1.0.0)
Using FirebaseCore (6.0.1)
Using FirebaseDatabase (6.0.0)
Using FirebaseInstanceID (4.1.0)
Using FirebaseMessaging (4.0.1)
Using FirebaseRemoteConfig (4.0.0)
Using FirebaseStorage (3.2.1)
Using FirebaseUI (7.0.0)
Using GoogleSignIn (4.4.0)
```

I'm referencing pods without a version specifier. Here's the relevant portion of my Podfile:
```
pod 'Firebase/Analytics'
pod 'Firebase/Auth'
pod 'Firebase/Core'
pod 'Firebase/Database'
pod 'Firebase/Messaging'
pod 'Firebase/RemoteConfig'
pod 'Firebase/Storage'
pod 'FirebaseUI/Auth'
pod 'FirebaseUI/Email'
pod 'FirebaseUI/Facebook'
pod 'FirebaseUI/Google'
```

jonathanfoster:
[...]

PR #686, opened 2019-05-26
Fix call to deprecated logInWithReadPermissions

jonathanfoster:
Fixes https://github.com/firebase/FirebaseUI-iOS/issues/684.


morganchen12:
Thanks for this PR. This should be released in master soon as well.

Issue #275, opened 2017-05-13
Add GitHub to supported IDPs


SUPERCILEX:
This is being tracked on Android (https://github.com/firebase/FirebaseUI-Android/issues/346) and is already available on the web.

samtstern:
This is done on Android:
https://github.com/firebase/FirebaseUI-Android/pull/1199

PR #702, opened 2019-06-07
Add github login to demo app

Yue-Wang-Google:


Yue-Wang-Google:
Will fix https://github.com/firebase/FirebaseUI-iOS/issues/275

Issue #464, opened 2018-06-07
iPad Email Sign In


dschlaba:
### Step 1: Are you in the right place?

  Yes. :)

### Step 2: Describe your environment

  * Objective C
  * iOS version: Deployment 10.0 / Devices 11.4
  * Firebase SDK version: 5.0.1
  * FirebaseUI version: 5.0.0
  * CocoaPods Version: 1.5.2
  
### Step 3: Describe the problem:

If the user selects Sign In with EMail, Name only appears if that e-mail address hasn't been used before. I'm assuming this is local to the app. However they can change the EMail address on the second screen where password is entered at which point no name is submitted to Google.

Maybe always display Name and populate prepopulate for known emails. Then if they change the email address on the second screen it would update. Also making name a required field would be a nice addition.

morganchen12:
This probably won't be a priority feature for us, but we're always happy to review a PR.

morganchen12:
Should be fixed in #757.

PR #757, opened 2019-08-23
Addressing issue #464

SebastienFCT:
Regarding the comments from the original ticket (https://github.com/firebase/FirebaseUI-iOS/issues/464) :

- Always displaying the `displayName field` hasn't been implemented as a custom field `_requireDisplayName` has been created specifically for this and should be used by the developer. 
- The email address field on the sign-in and sign-up controllers have been disabled from editing for the following reason: 
    - the email address is being verified in the previous screen and allowing the user to edit it at this point bypasses all the previous validations (so the user can potentially signup with an existing email address for example). I tested all cases and it does throw an error message properly handled every time but it still feels like a bug and potentially opens the door to other issue for future development. The back button provides a mean to modify the email address **with** all required validations

Let me know if that makes sense or if you'd like to see any other change applied and I'll update the PR


renkelvin:
Looks good to me. Thanks!

PR #762, opened 2019-08-28
Support dark mode in auth flow

### LINKED ISSUE #723: Dark mode
* Swift 5
  * iOS 13 beta 3
  * Firebase SDK version: 6.3.0
  * FirebaseUI version: 8.0.2
  * CocoaPods Version: 1.7.3
 
#### Steps to reproduce:

  1. Authentication views of phone sign in and email-password sign in from the prebuilt UI (FirebaseUI)
  
#### Observed Results:

  * Color of the label texts (“Email”, “Password”) and background in sign-in/sign-up view do not change based on user’s dark mode setting.  
* The reason is those UI elements have color explicitly assigned. 
  
#### Expected Results:

  * It would be great if the prebuilt UI can adapt to the system-wide dark mode feature of the iOS 13. I attempted to change the colours. Since I am not quite familiar with ObjC so I could change only the password label into what you can see in the attached screenshot. The black “email” label and white background were already there in the prebuilt UI
D1FFB1FE-C837-449A-9535-BD7689C04C1E

cbess:
Resolves #723 for the Swift code authentication flow.

### DIFF
--- a/Auth/FirebaseAuthUI/FUIAuthTableViewCell.m
+++ b/Auth/FirebaseAuthUI/FUIAuthTableViewCell.m
@@ -18,6 +18,15 @@
 
 @implementation FUIAuthTableViewCell
 
+- (void)awakeFromNib {
+  [super awakeFromNib];
+  
+  if (@available(iOS 13.0, *)) {
+    self.textField.textColor = [UIColor labelColor];
+    self.label.textColor = [UIColor labelColor];
+  }
+}
+
 - (void)setLabel:(UILabel *)label {
   _label = label;
   [self layoutIfNeeded];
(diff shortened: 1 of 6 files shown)


cbess:
I'll check and let ya know

cbess:
> LGTM. Does Xcode 10.x support the new key/values in the sample storyboard file?

So, it opens with no problem. But, some of the constraints and such render oddly. But, it won't build due to the `[UIColor labelColor]` use. Essentially, due to the iOS 13 SDK not being available in Xcode 10. Thoughts?

cbess:
I suppose I could use `#ifdef __IPHONE_13_0`

morganchen12:
No, no need to update the code. Let's keep this PR open until we update Xcode on Travis to support building the new IB file.

melsam:
Is there an ETA to get this merged? Now that iOS 13 has officially launched, not having this is preventing our app from fully supporting dark mode.

cbess:
@morganchen12 I'm also curious about when it will be merged, now that Xcode 11 is generally available via the Mac App Store.

morganchen12:
I plan on merging this week. Bumping the dependency in Travis should take less than a day.

morganchen12:
Thanks @cbess for contributing!

PR #786, opened 2019-10-03
Authentication with e-mail link was not working in iOS13.

### LINKED ISSUE #785: iOS 13 Email Link Sign In doesn't work
* Objective C or Swift: Swift
  * iOS version: 13.1.1
  * Firebase SDK version: 6.9.0
  * FirebaseUI version: 8.1.0
  * CocoaPods Version: 1.8.1

### Step 3: Describe the problem:

The “Signed In!” alert disappears automatically, so `FUIAuthDelegate` is not called.

On iOS 13, `UIWindow` reference seems to be lost.

#### Steps to reproduce:

  1. [Configure Email Link Sign In](https://github.com/firebase/FirebaseUI-iOS/tree/master/Auth#configuring-email-link-sign-in)
  1. Email link authentication

イメージ

#### Observed Results:

  * Signed In! Alert disappears automatically
  * Authentication seems to be complete, but `FUIAuthDelegate` is not called

#### Expected Results:

  * In iOS 12, Signed In! Alert disappears when user taps OK
  * In iOS 12, `FUIAuthDelegate` was called

#### Relevant Code:

My fixes
https://github.com/wata/FirebaseUI-iOS/commit/f3b757be936dde6614b2104679cdcd77ebeb5040

SO
- https://stackoverflow.com/questions/58131996/every-uialertcontroller-disappear-automatically-before-user-responds-since-ios
- https://stackoverflow.com/questions/58148334/uialertcontroller-disappearing-since-ios-13

rojo69:
The success alert got removed right away, without notifying the delegate.

Fixes #785

### DIFF
--- a/Auth/FirebaseAuthUI/FUIAuthBaseViewController.m
+++ b/Auth/FirebaseAuthUI/FUIAuthBaseViewController.m
@@ -294,8 +294,8 @@ static NSString *const kAuthUICodingKey = @"authUI";
             Earlier iOS versions established a strong reference to the window when makeKeyAndVisible was called.
             Now we add one from the alert controller, to prevent objects from getting garbage collected right away.
         */
-        static void *key;
-        objc_setAssociatedObject(alertController, key, window, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+        static char key;
+        objc_setAssociatedObject(alertController, &key, window, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
 	  }
   }
 }
--- a/Auth/FirebaseAuthUI/FUIAuthBaseViewController.m
+++ b/Auth/FirebaseAuthUI/FUIAuthBaseViewController.m
@@ -21,6 +21,7 @@
 #import "FUIAuthStrings.h"
 #import "FUIAuthUtils.h"
 #import "FUIAuth_Internal.h"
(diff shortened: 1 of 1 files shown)


Auth/FirebaseAuthUI/FUIAuthBaseViewController.m
```diff
@@ -287,6 +288,15 @@ + (void)showAlertWithTitle:(nullable NSString *)title
...
     [window makeKeyAndVisible];
     [viewController presentViewController:alertController animated:YES completion:nil];
+	  
+    if (@available(iOS 13.0, *)) {
+        /*
+            Earlier iOS versions established a strong reference to the window when makeKeyAndVisible was called.
+            Now we add one from the alert controller, to prevent objects from getting garbage collected right away.
+        */
+        static void *key;
+        objc_setAssociatedObject(alertController, key, window, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
```
morganchen12: I think using `key` here is technically undefined behavior since `key` is uninitialized. Consider using a selector instead, or setting `key` to `nil` and using `&key`.

rojo69:
Yes, you are correct. Will change it to the address of a static char.

morganchen12:
Thanks!

PR #824, opened 2020-01-08
Fix auto upgrade for oauth provider

renkelvin:
Fix https://github.com/firebase/firebase-ios-sdk/issues/4620

### DIFF
--- a/Auth/FirebaseAuthUI/FUIAuth.m
+++ b/Auth/FirebaseAuthUI/FUIAuth.m
@@ -238,9 +238,10 @@ static NSString *const kFirebaseAuthUIFrameworkMarker = @"FirebaseUI-iOS";
       // Check for "credential in use" conflict error and handle appropriately.
       if (error.code == FIRAuthErrorCodeCredentialAlreadyInUse) {
         FIRAuthCredential *newCredential = error.userInfo[FIRAuthErrorUserInfoUpdatedCredentialKey];
-        NSDictionary *userInfo = @{
-          FUIAuthCredentialKey : newCredential,
-        };
+        NSDictionary *userInfo = @{ };
+        if (newCredential) {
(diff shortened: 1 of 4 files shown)


morganchen12:
For Sign in with Apple, don't we need the new user credential from the `FIRAuthErrorUserInfoUpdatedCredentialKey` key?

renkelvin:
It's already handled: https://github.com/firebase/FirebaseUI-iOS/blob/6d6dfd6bb0ab36442d34fff6aea82729345009ab/Auth/FirebaseAuthUI/FUIAuth.m#L244

renkelvin:
Hmm, it seems limited to Phone auth only. Let me fix it.

Auth/FirebaseAuthUI/FUIAuth.m
```diff
@@ -237,12 +237,7 @@ - (void)autoUpgradeAccountWithProviderUI:(id<FUIAuthProvider>)providerUI
     if (error) {
       // Check for "credential in use" conflict error and handle appropriately.
       if (error.code == FIRAuthErrorCodeCredentialAlreadyInUse) {
-        FIRAuthCredential *newCredential = credential;
-        // Check for and handle special case for Phone Auth Provider.
-        if (providerUI.providerID == FIRPhoneAuthProviderID) {
-          // Obtain temporary Phone Auth credential.
-          newCredential = error.userInfo[FIRAuthErrorUserInfoUpdatedCredentialKey];
-        }
+        FIRAuthCredential *newCredential = error.userInfo[FIRAuthErrorUserInfoUpdatedCredentialKey];
```
morganchen12: This implementation seems to be provider-specific, so we should check for nil here to avoid crashing on line 242 below.

renkelvin: Yeah, good catch.

PR #829, opened 2020-01-16
Center align sign in with Apple button

renkelvin:


Auth/FirebaseAuthUI/FUIAuthProvider.h
```diff
@@ -48,6 +48,15 @@ typedef void (^FUIAuthProviderSignInCompletionBlock) (
  */
 typedef NSString *FUIAuthProviderSignInUserInfoKey NS_TYPED_ENUM;
 
+/**
+   @typedef FUIButtonAlignment
+   @brief The alignment of the icon and text of the button.
+*/
+typedef NS_ENUM(NSInteger, FUIButtonAlignment) {
+    FUIButtonAlignmentLeft,
```
morganchen12: This should be renamed to `FUIButtonAlignmentLeading`, since it adheres to the leading edge (on the right for RTL languages, on the left otherwise).

PR #841, opened 2020-03-07
Update StorageUI integration with SDWebImage

dreampiggy:

This PR add the following features:

1. Use [SDWebImage 5.6.0](https://github.com/SDWebImage/SDWebImage/releases/tag/5.6.0) 's loader context option, instead of dummy manager instance. This is easier to use. And can avoid user who overrde the default manager's property (like transformer), which cause the undefined behavior when using with FirebaseUI Storage
2. Add the support to load `gs://` URl directly, not always need to create a wrapper `FIRStorageReference`.
3. Add the support to progressive downloading and decoding of images, this can be the use case who want to load partial images and display for users.

The dependency of SDWebImage bumped to 5.6.0, beucase of the PR feature 1 need the new API.

CC @morganchen12

### DIFF
--- a/FirebaseUI.podspec
+++ b/FirebaseUI.podspec
@@ -37,7 +37,7 @@ Pod::Spec.new do |s|
     storage.public_header_files = 'Storage/FirebaseStorageUI/*.h'
     storage.source_files = 'Storage/FirebaseStorageUI/*.{h,m}'
     storage.dependency 'Firebase/Storage', '~> 6.0'
-    storage.dependency 'SDWebImage', '~> 5.0'
+    storage.dependency 'SDWebImage', '~> 5.6'
     storage.xcconfig = { 'HEADER_SEARCH_PATHS' => '$(PODS_ROOT)/FirebaseUI/FirebaseStorageUI' }
   end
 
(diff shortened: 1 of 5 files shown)


dreampiggy:
@morganchen12 Hi.

After I update the demo to try the code, I found the strange exception thrown from Storage.

image

However, the official documentation told me it's OK to pass the Full URI into the `storageWithURL:`

image


https://firebase.google.com/docs/storage/ios/create-reference

Which is correct ? Is this exception the designed behavior ? And whatever, this need to be fixed (I need another PR)

dreampiggy:
If I can not init a full URI into the `FIRStorageRef` directly, what can I do now ? Should I parse the URL path and use this arg to call `child:` ?

morganchen12:
@dreampiggy the reason the crash happens is because you're trying to initialize a bucket with a full path to a specific file. The bucket is just the host (as you've found out).

I'm not sure why we don't allow creating a reference without first initializing the bucket.

dreampiggy:
> However, the official documentation told me it's OK to pass the Full URI into the storageWithURL:

Seems we need to update the documentation there ? It cause misunderstanding, because actually, you can not write code like this:

```swift
// This is equivalent to creating the full reference
let storagePath = "\(your_firebase_storage_bucket)/images/space.jpg"
spaceRef = storage.reference(forURL: storagePath)    
```

This will cause crash on exception, and worse, Swift can not even handle Objective-C's NSException, dangerous.

dreampiggy:
Seems the code comes from the Day 1 of FirebaseStorage 😭 

https://github.com/firebase/firebase-ios-sdk/blame/master/FirebaseStorage/Sources/FIRStorage.m#L132

The documentation does not write correctly for that `storageWithURL:` behavior.

morganchen12:
I'll see what we can do to change the behavior of Firebase Storage.

PR #842, opened 2020-03-10
Fix StorageUI Bug and Update demo usage for SDWebImgae integration

dreampiggy:

Fix issues related StorageUI and SDWebImage, introduced in #841 

 1. The FIRStorage throw exception on passed url contains object path
2. The progress block may been called on main queue, should use another coder queue for decoding
3. The expectedSize should use URLResponse's info to check, but not the bodyLength

CC @morganchen12

### DIFF
--- a/Storage/FirebaseStorageUI/FUIStorageImageLoader.m
+++ b/Storage/FirebaseStorageUI/FUIStorageImageLoader.m
@@ -31,6 +31,12 @@
 
 @end
 
+@interface FUIStorageImageLoader ()
+
+@property (nonatomic, strong) dispatch_queue_t coderQueue;
+
+@end
+
 @implementation FUIStorageImageLoader
 
 + (FUIStorageImageLoader *)sharedLoader {
@@ -45,7 +51,8 @@
 - (instancetype)init {
   self = [super init];
   if (self) {
-    self.defaultMaxImageSize = 10e6;
+    _defaultMaxImageSize = 10e6;
+    _coderQueue = dispatch_queue_create("com.google.firebaseui.storage.coderQueue", DISPATCH_QUEUE_SERIAL);
   }
   return self;
 }
@@ -66,8 +73,9 @@
   FIRStorageReference *storageRef = url.sd_storageReference;
   if (!storageRef) {
     // Create Storage Reference from URL
-    FIRStorage *storage = [FIRStorage storageWithURL:url.absoluteString];
-    storageRef = storage.reference;
+    NSString *bucketUrl = [NSString stringWithFormat:@"gs://%@", url.host];
+    FIRStorage *storage = [FIRStorage storageWithURL:bucketUrl];
+    storageRef = [storage referenceWithPath:url.path];
     url.sd_storageReference = storageRef;
   }
   
@@ -97,13 +105,15 @@
                                                            return;
                                                          }
                                                          // Decode the image with data
-                                                         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{
-                                                           UIImage *image = SDImageLoaderDecodeImageData(data, url, options, context);
-                                                           dispatch_main_async_safe(^{
-                                                             if (completedBlock) {
-                                                               completedBlock(image, data, nil, YES);
-                                                             }
-                                                           });
+                                                         dispatch_async(self.coderQueue, ^{
+                                                           @autoreleasepool {
+                                                             UIImage *image = SDImageLoaderDecodeImageData(data, url, options, context);
+                                                             dispatch_main_async_safe(^{
+                                                               if (completedBlock) {
+                                                                 completedBlock(image, data, nil, YES);
+                                                               }
+                                                             });
+                                                           }
                                                          });
                                                        }];
   // Observe the progress changes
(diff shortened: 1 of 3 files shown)


Storage/FirebaseStorageUI/FUIStorageImageLoader.m
```diff
@@ -45,7 +51,8 @@ + (FUIStorageImageLoader *)sharedLoader {
 - (instancetype)init {
   self = [super init];
   if (self) {
-    self.defaultMaxImageSize = 10e6;
+    _defaultMaxImageSize = 10e6;
+    _coderQueue = dispatch_queue_create("com.google.firebaseui.storage.coderQueue", DISPATCH_QUEUE_SERIAL);
```
dreampiggy: Should we use a Loader (shared instance) level queue ?
Or, should we use per-download-task level queue ?

I'm afraid, if there are 10 concurrent download in the same time, this serial queue will been busy and block each download task.

dreampiggy: Implements the logic to use per-download-task level decoding queue.

This is the current suitable solution for this case, which can ensure callback order during progressive decoding (later appending data will neven produce image before previous one)

samples/swift/FirebaseUI-demo-swift/Samples/StorageViewController.swift
```diff
@@ -50,12 +54,19 @@ class StorageViewController: UIViewController {
...
+                               maxImageSize: 10000000,
+                               placeholderImage: nil,
+                               options: [.progressiveLoad]) { (image, error, cacheType, storageRef) in
       if let error = error {
         print("Error loading image: \(error)")
       }
     }
   }
+  
+  @objc fileprivate func flushCache() {
```
morganchen12: nit: this method can probably be private instead of fileprivate.

Storage/FirebaseStorageUI/FUIStorageImageLoader.m
```diff
@@ -85,7 +86,8 @@ - (BOOL)canRequestImageForURL:(NSURL *)url {
     size = self.defaultMaxImageSize;
   }
   // Download the image from Firebase Storage
-  
+  // Each download task use independent serial coder queue, to ensure callback in order during prorgessive decoding
+  dispatch_queue_t coderQueue = dispatch_queue_create("com.google.firebaseui.storage.coderQueue", DISPATCH_QUEUE_SERIAL);
```
morganchen12: 👍 

Maybe name this `decoderQueue`, since it's where all the image decoding happens. If you want to cancel previous decoding operations when a new operation is enqueued, you can use `NSOperationQueue` instead of a dispatch queue.

samples/swift/FirebaseUI-demo-swift/Samples/StorageViewController.swift
dreampiggy: Seems other target action method use fileprivate (like the above `loadButtonPressed`) as well 😅 Change them at the same time ?

Storage/FirebaseStorageUI/FUIStorageImageLoader.m
dreampiggy: Using `NSOperationQueue` is a choice if we can cancel all previous progressive decoding when FIRStorageDownloadTask is cancelled. Need some extra check code for `FIRStorageDownloadTask`'s status.

Is there public API for this ? I found one `FIRStorageDownloadTask.state` (in `FIRStorageTask_Private.h` 😓 ), which is a enum. Need we check the state and earily return in the operation block body ?

dreampiggy: I've done using NSOperationQueue, without checking the `FIRStorageDownloadTask.state`.  The reason:

1. Change the behavior, if we've already have decoding process, we don't decode current partial images. Means the there are always 0 or 1 task been executed, no queue to wait. Directlly ignore the dummy decoding. (Because user only need one picture at the time, right ?)

2. When `FIRStorageDownloadTask` get cancelled, they will no longer report progress callback, so we don't need extra check inside NSBlockOperation, just let it done.

This design is the same as SDWebImage itself, see: https://github.com/SDWebImage/SDWebImage/pull/2483

dreampiggy: image

samples/swift/FirebaseUI-demo-swift/Samples/StorageViewController.swift
morganchen12: You can leave this as-is then, I will address them separately later.

dreampiggy: Done.

Actually in daily Swift demo code, I usually leave it as the default `func`, which is internel level control. Isn't that suitable ?

Storage/FirebaseStorageUI/FUIStorageImageLoader.m
```diff
@@ -85,27 +86,29 @@ - (BOOL)canRequestImageForURL:(NSURL *)url {
...
-                                                           UIImage *image = SDImageLoaderDecodeImageData(data, url, options, context);
-                                                           dispatch_main_async_safe(^{
-                                                             if (completedBlock) {
-                                                               completedBlock(image, data, nil, YES);
-                                                             }
-                                                           });
-                                                         });
-                                                       }];
+  // Each download task use independent serial coder queue, to ensure callback in order during prorgessive decoding
+  NSOperationQueue *coderQueue = [NSOperationQueue new];
```
morganchen12: It looks like a new `coderQueue` is allocated on each invocation of this method, but never retained anywhere. Is this intentional?

Storage/FirebaseStorageUI/FUIStorageImageLoader.m
morganchen12: Thanks for looking into this!

Storage/FirebaseStorageUI/FUIStorageImageLoader.m
dreampiggy: Bug seems I reset the wrong code. The dispatch_async should be replaced with coderQueue addOperationWithBlock:...Wait me fix this

dreampiggy: The lifetime is associated to the completion block, and the completion block lifetime is associated to `FIRStorageDownloadTask`. This implements the design I want : Each download task use independent serial coder queue

morganchen12: Sounds good.

dreampiggy: Or, if you think it's hard to find, we can change this into an associated object of `FIRStorageDownloadTask`, like

```
@interface FIRStorageDownloadTask (StorageUI)
@property (nonatomic) NSOperationQueue *coderQueue;
@end
@implementation FIRStorageDownloadTask (StorageUI)
- (NSOperationQueue *)coderQueue {
    NSOperationQueue *coderQueue = objc_getAsscoaitedObject(self, @selector(coderQueue));
    if (!coderQueue) {
        coderQueue = [NSOperationQueue new];
        coderQueue.maxConcurrentOperationCount = 1;
        objc_setAssociatedObject(self, @selector(coderQueue), coderQueue, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
    } 
    return coderQueue;
}
@end
```

morganchen12: I'm ok with the way the code is, just wanted to clarify 👍

dreampiggy:
If you find there are no extra review suggestion to check, feel free to merge. :)

dreampiggy:
The Travis-CI failed bacause of CocoaPods repo update failed :(

@morganchen12 Can you update the Gemfile and CI to use new CocoaPods version ? At least v1.8.0, which introduce the CDN trunk instead of Spec repo. Which don't need to wasting time to git clone the https://github.com/CocoaPods/Specs. It use a CDN for HTTP download, 10x faster.

morganchen12:
Updated gems in #844.

morganchen12:
The changes look fine to me, I'll take care of fixing the build. Thanks @dreampiggy!

Issue #885, opened 2020-07-22
FirebaseUI Facebook Auth Dependency Update


wimbledon:
Please update [FirebaseUI.podspec](https://github.com/firebase/FirebaseUI-iOS/blob/master/FirebaseUI.podspec) `FirebaseUI/Facebook` dependency from `'FBSDKLoginKit', '~> 5.0'` to `'FBSDKLoginKit', '~> 7.0'`.

`FBSDKLoginKit ~> 5.0` is causing  crash issues like https://stackoverflow.com/questions/62965382 and other SDKs depend on the latest version of [Facebook iOS SDK](https://github.com/facebook/facebook-ios-sdk/releases).

wimbledon:
@morganchen12, any update on this issue?

morganchen12:
I can update this today, but it will constitute a breaking change.

morganchen12:
This is fixed in FirebaseUI 9.

wimbledon:
Thank you, @morganchen12.

PR #887, opened 2020-07-29
Prepare for major version bump

morganchen12:
Some cleanup to address #885.

### DIFF
--- a/AnonymousAuth/Podfile
+++ b/AnonymousAuth/Podfile
@@ -1,4 +1,4 @@
-platform :ios, '8.0'
+platform :ios, '9.0'
 
 target 'FirebaseAnonymousAuthUI' do
   use_frameworks!
--- a/FacebookAuth/Podfile
+++ b/FacebookAuth/Podfile
@@ -1,7 +1,7 @@
 # source 'sso://cpdc-internal/firebase'
 source 'https://github.com/CocoaPods/Specs.git'
 
-platform :ios, '8.0'
+platform :ios, '9.0'
 
 target 'FirebaseFacebookAuthUI' do
   use_frameworks!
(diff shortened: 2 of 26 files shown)


rosalyntan:
Adding @renkelvin as I'm not really familiar with the project.pbxproj updates.

morganchen12:
Yes--this will result in a major (breaking) change.

FacebookAuth/Podfile
```diff
@@ -1,18 +1,19 @@
...
   # The facebook pods generate an unreasonable number of warnings.
   inhibit_all_warnings!
 
-  pod 'FBSDKLoginKit', '~> 5.0'
-  pod 'FBSDKCoreKit', '~> 5.0'
+  pod 'FBSDKLoginKit', '~> 7.0'
+  pod 'FBSDKCoreKit'
 
   pod 'FirebaseUI/Auth', :path => '../'
+  pod 'PromisesObjC'
```
renkelvin, line 16: Is this required?

AnonymousAuth/Podfile
```diff
@@ -1,4 +1,4 @@
-platform :ios, '8.0'
+platform :ios, '9.0'
```
renkelvin, line 1: the core sdk currently support ios 8. If it's going to be bumped to 9.0(most likely), this is fine. Otherwise, we may want to keep FirebaseUI consistent with the core sdk.

morganchen12: Yeah, I mentioned this to Paul and he said bumping to 9.0 makes sense.

FacebookAuth/Podfile
morganchen12: No, I'll remove it

AnonymousAuth/Podfile
renkelvin: Sounds good.

paulb777: we ended up bumping to 10 for Firebase 7. Probably should do that here too.

morganchen12: Done.

PR #907, opened 2020-10-10
Fix Apple login button color to white when in dark mode

cozzin:
This PR fix the issue related Apple login button color, introduced in #841

We should set white button color when background is dark or colored.
https://developer.apple.com/design/human-interface-guidelines/sign-in-with-apple/overview/buttons/
image


morganchen12 (approved):
Thanks!

carloshwa:
Is there a way to make this configurable? The background of my login screen is a dark color, in both light and dark modes, so I need the button to be white in both cases.

cozzin:
Hi, @carloshwa. You can check new method in https://github.com/firebase/FirebaseUI-iOS/pull/908

PR #908, opened 2020-10-18
Support setting custom UserInterfaceStyle for AppleAuth

cozzin:
This PR is related with https://github.com/firebase/FirebaseUI-iOS/pull/907#issuecomment-711027069.
Depending on the situation, it is need to configure custom `UserInterfaceStyle`.

### DIFF
--- a/OAuth/FirebaseOAuthUI/FUIOAuth.m
+++ b/OAuth/FirebaseOAuthUI/FUIOAuth.m
@@ -191,11 +191,15 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 + (FUIOAuth *)appleAuthProvider {
+  return [self appleAuthProviderWithUserInterfaceStyle:UITraitCollection.currentTraitCollection.userInterfaceStyle];
+}
+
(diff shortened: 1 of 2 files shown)


OAuth/FirebaseOAuthUI/FUIOAuth.m
```diff
@@ -191,11 +191,15 @@ + (FUIOAuth *)yahooAuthProvider {
 }
 
 + (FUIOAuth *)appleAuthProvider {
+  return [self appleAuthProviderWithUserInterfaceStyle:UITraitCollection.currentTraitCollection.userInterfaceStyle];
```
morganchen12, line 194: Does this need an availability check?

cozzin: Hi, @morganchen12.   
Because we are using [`ASAuthorizationAppleIDRequest`](https://developer.apple.com/documentation/authenticationservices/asauthorizationappleidrequest?language=objc) internally, it needs availability check.

https://github.com/firebase/FirebaseUI-iOS/blob/5e7e9bcb6b6674effdd5e8cc0de4b138bdfd5d7c/OAuth/FirebaseOAuthUI/FUIOAuth.m#L255-L263

morganchen12 (approved):
Thanks!

PR #909, opened 2020-10-18
Update FirebaseUI version for sample projects

cozzin:
* Update minimum target version to iOS 10.0
* Update FirebaseUI version to 9.0.0


morganchen12:
Thanks @cozzin, can you run `pod deintegrate` on the samples and then commit the result?

cozzin:
Thank you @morganchen12, please check 81b49f0.

morganchen12 (approved):
Thanks!

PR #911, opened 2020-10-28
Fix issue where CI passes even though tests fail.

rosalyntan:
Fix issue where CI passes even though tests fail, add OAuth tests to CI, and fix failing tests.


berjayasompo (approved):
Tq

berjayasompo (approved):

morganchen12 (approved):
LGTM on green

PR #912, opened 2020-10-29
Fixes wrong selector for check

chbeer:
Fixes checking for selector when calling delegate.


morganchen12 (approved):
Thanks!

chbeer:
@googlebot I signed it!

PR #913, opened 2020-10-29
Try again to fix Travis CI.

rosalyntan:
#911 fixed the broken tests, but the Travis CI still has the underlying issue where it passes even when there are failing tests.


morganchen12 (approved):

rosalyntan:
Confirmed that Travis CI correctly fails now when there are failing tests: https://travis-ci.org/github/firebase/FirebaseUI-iOS/builds/739974734

PR #942, opened 2021-04-02
Add Swift Package Manager support

morganchen12:
Addresses https://github.com/firebase/FirebaseUI-iOS/issues/886, but lacks support for Facebook and Google auth.


EmailAuth/FirebaseEmailAuthUI/FUIEmailAuth.m
```diff
@@ -20,7 +20,11 @@
 #import <FirebaseCore/FIROptions.h>
 #import <FirebaseAuth/FIRAuth.h>
 #import <FirebaseAuth/FirebaseAuth.h>
+#if SWIFT_PACKAGE
```
paulb777, line 23: This `#if` is no longer necessary and `#import <GoogleUtilities/GULUserDefaults.h>` should work for both SPM and CocoaPods.

Package.swift
```diff
@@ -0,0 +1,253 @@
...
+      exclude: ["Info.plist"],
+      resources: [
+        .process("Resources"),
+      ],
+      publicHeadersPath: ".",
+      cSettings: [
+        .headerSearchPath("./"),
+      ]
+    ),
+    // Facebook doesn't seem to vend their ObjC libraries through SPM, though thier
```
paulb777, line 151: s/thier/their/

paulb777 (approved):
Can you follow https://github.com/firebase/firebase-ios-sdk/blob/master/HeadersImports.md to reduce some of the SWIFT_PACKAGE conditionals? Specifically, nest all public headers in `{target}/Public/FirebaseUI` so that the CocoaPods and SPM imports can match.

I'm fine to merge if this is functional and you want to defer these cleanups

morganchen12:
Would following HeadersImports require moving to a `pod gen`-based workflow as well? If so, I'll likely leave it out of this PR.

paulb777:
The location of the public headers and how they're referenced shouldn't have anything to do with whether a `pod gen` workflow is used or not.

morganchen12:
I'm now running into [this issue](https://forums.swift.org/t/unable-to-compile-swift-package-when-used-in-xcode/36081), so unifying the header imports is the only way forward for now.

morganchen12:
Should be ready for final review. I can't figure out why my spm test project won't build unless I use module imports for FirebaseUI, but Firebase and all its dependencies work fine with angle bracket header imports.

paulb777:
GoogleUtilities like https://github.com/google/GoogleUtilities/tree/main/GoogleUtilities/Environment/Public/GoogleUtilities is a good example of how public headers should be structured to support multiple targets in a single repo so they work with both CocoaPods and SPM

PR #968, opened 2021-06-22
Remove unnecessary dependencies on Firebase umbrella podspec

paulb777:
This is an alternative fix to https://github.com/firebase/firebase-ios-sdk/pull/8284 for b/191680865. More testing is needed and I'll investigate if there are other published podspecs depending on the Firebase pod.


maksymmalyhin (approved):
LGTM

morganchen12 (approved):

PR #994, opened 2021-07-26
Update English strings with less ambiguous message

morganchen12:


rosalyntan (approved):
Do we need to update any string translations as well?

morganchen12:
Yep, I can take care of the translation console and generate another PR with the updated languages.

PR #996, opened 2021-07-28
Attach a nonce to Sign in with Apple requests.

rosalyntan:


FirebaseAuthUI/Sources/FUIAuthUtils.m
```diff
@@ -74,4 +76,47 @@ + (nullable UIImage *)imageNamed:(NSString *)name fromBundle:(nullable NSBundle
   }
 }
 
++ (NSString *)randomNonce {
```
morganchen12, line 79: wdyt about adding a comment saying this method is adapted from the [auth0 docs](https://auth0.com/docs/authorization/mitigate-replay-attacks-when-using-the-implicit-flow#generate-a-cryptographically-random-nonce)? I'm skeptical of my own ability to write any crypto code, so it gives me confidence to say this was ported from a more reputable source.

FirebaseAuthUI/Sources/FUIAuthUtils.m
```diff
@@ -74,4 +76,47 @@ + (nullable UIImage *)imageNamed:(NSString *)name fromBundle:(nullable NSBundle
...
+  NSString *characterSet = @"0123456789ABCDEFGHIJKLMNOPQRSTUVXYZabcdefghijklmnopqrstuvwxyz-._";
+  NSMutableString *result = [NSMutableString string];
+  NSInteger remainingLength = 32;
+
+  while (remainingLength > 0) {
+    NSMutableArray *randoms = [NSMutableArray arrayWithCapacity:16];
+    for (NSInteger i = 0; i < 16; i++) {
+      uint8_t random = 0;
+      int errorCode = SecRandomCopyBytes(kSecRandomDefault, 1, &random);
+      NSAssert(errorCode == errSecSuccess, @"Unable to generate nonce: OSStatus %i", errorCode);
```
morganchen12, line 89: If this line fails in CI again with an `unused variable` error, what's probably happening is the assert is being compiled out in release mode. We should change this to manually constructing and throwing an `NSException` instead since we'd never want this code to continue execution in a release build.

morganchen12 (commented):
Thanks for the cleanup!

morganchen12 (approved):
LGTM

b/194944488

FirebaseAuthUI/Sources/FUIAuthUtils.m
rosalyntan: Done.

PR #1010, opened 2021-11-08
Should callback Apple's error status instead of doing nothing (canceled/failed/invalid/Nothandled/Unknown)

for-meng:
When I log in with Apple and I have been waiting for a callback, but due to some internal Apple reasons, there has been no callback, which caused me to be unable to determine what happened.


for-meng:
> @googlebot I signed it!

for-meng:
@googlebot I fixed it.

PR #1011, opened 2021-11-08
Should callback Apple's error status instead of doing nothing (canceled/failed/invalid/Nothandled/Unknown)

for-meng:
When I log in with Apple, I have been waiting for a callback, but due to some errors within Apple, there has been no callback, so I can't determine what happened. So I think the error generated by apple should be called back here so that the developer knows what happened.


morganchen12 (approved):
Thanks!

PR #1023, opened 2022-04-08
[Firebase 9] Add spec `swift_version`

ncooke3:
Come Firebase 9, all SDKs that depend on Firebase will need to specify a `swift_version`.

I've set the `swift_version` of all specs to match that of Firebase.

#no-changelog


paulb777:
We might want to release this along with updating the Firebase deps to Firebase 9 and release immediately after Firebase 9 releases.

PR #1024, opened 2022-04-13
Update FirebaseUI for v8.x and v9.x compatibility

paulb777:
podspec and source updates to support both Firebase 8.x and 9.x

It uses `pre` prefixed tags for testing on SpecsStaging.


paulb777:
Changes look good. If you do a merge, don't squash since we're depending on the previous commit for pre-v9 Firebase testing

franckbleyer1 (commented):
Ever done;-)

Issue #1111, opened 2022-10-21
gtm-session-fetcher issue with firebase version 10.0 SDK


ian-sayles:
* Objective C or Swift: Swift
  * iOS version: n/a
  * Firebase SDK version: 10.0
  * FirebaseUI version: 12.3
  * CocoaPods Version: n/a using Package Manager
  
When upgrading Firebase SDK from 9.6 I get the following issue:

'firebaseui-ios' is incompatible with 'firebase-ios-sdk' because 'firebase-ios-sdk' 10.0.0 depends on 'gtm-session-fetcher' 2.1.0..<3.0.0 and no versions of 'firebase-ios-sdk' match the requirement 10.0.1..<11.0.0.

'firebaseui-ios' >= 12.0.0 practically depends on 'gtm-session-fetcher' 1.4.0..<2.0.0 because 'firebaseui-ios' 12.2.0 depends on 'gtm-session-fetcher' 1.4.0..<2.0.0 and 'firebaseui-ios' 12.3.0 depends on 'gtm-session-fetcher' 1.4.0..<2.0.0.

'firebaseui-ios' {12.0.0..<12.2.0, 12.2.1..<12.3.0, 12.3.1..<13.0.0} practically depends on 'gtm-session-fetcher' 1.4.0..<2.0.0 because 'firebaseui-ios' 12.1.0 depends on 'gtm-session-fetcher' 1.4.0..<2.0.0 and 'firebaseui-ios' 12.1.1 depends on 'gtm-session-fetcher' 

Is this issue due to my setup or Firebaseui is not yet compatible with latest Firebase SDK?

ArturoLee:
I'n getting the same issue pretty much. I've been trying ti reset cache and derived data but nothing has worked.

Xcode 14.0
error project: Failed to resolve dependencies Dependencies could not be resolved because root depends on 'firebase-ios-sdk' 10.0.0..<11.0.0 and root depends on 'firebaseui-ios' 12.3.0..<13.0.0.
'firebaseui-ios' is incompatible with 'firebase-ios-sdk' because 'firebase-ios-sdk' 10.0.0 depends on 'gtm-session-fetcher' 2.1.0..<3.0.0 and no versions of 'firebase-ios-sdk' match the requirement 10.0.1..<11.0.0.
[...]

PR #1112, opened 2022-10-23
Updated GTMSessionFetcher package version

ArturoLee:
The GTMSessionFetcher package version now matches the one in Firebase-iOS-SDK. #1111

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -87,7 +87,7 @@ let package = Package(
     .package(
       name: "GTMSessionFetcher",
       url: "https://github.com/google/gtm-session-fetcher.git",
-      "1.4.0" ..< "2.0.0"
+      "2.1.0" ..< "3.0.0"
     ),
     .package(
       name: "SDWebImage",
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -87,7 +87,7 @@ let package = Package(
     .package(
       name: "GTMSessionFetcher",
       url: "https://github.com/google/gtm-session-fetcher.git",
-      "1.4.0" ..< "2.0.0"
+      "2.1.0" ..< "3.0.0"
```
paulb777, line 90: May want to be `"1.4.0" ..< "3.0.0"` to work with both Firebase 9 and 10.

ArturoLee: I could change it to that. I just matched how it currently is in the iOS SDK: https://github.com/firebase/firebase-ios-sdk/blob/master/Package.swift

morganchen12: Please make it `"1.4.0" ..< "3.0.0"` so we can release it in a non-breaking change.

morganchen12:
Thanks!

ArturoLee:
@morganchen12 #1111  can be closed now. (didn't link the issue correctly)

PR #1177, opened 2023-12-21
rename master to main

morganchen12:


morganchen12:
pod spec lint warnings will be fixed in a follow-up.

PR #1188, opened 2024-05-14
[Bug] Fix firebase-ios-sdk/#12960

ncooke3:
#no-changelog


ncooke3:
cc: @morganchen12 for next steps on releasing

ncooke3:
Tested without fix -> error
With fix -> no error

PR #1193, opened 2024-07-23
More Firebase 11 updates

paulb777:
Build from SPM with both Firebase 10 and Firebase 11

Migrate to GoogleSignIn 7. I disabled and left a TODO for the unit tests that need to be updated in ` FirebaseGoogleAuthUITests/FirebaseGoogleAuthUITests.m`

There are several ugly string changes because SPM does not include the libraries global string constants. (CocoaPods does). 

Once the minimum dependency is Firebase 11, they can be updated to the Swift symbol.

### DIFF
--- a/FirebaseAuthUI/Sources/FUIAccountSettingsOperationDeleteAccount.m
+++ b/FirebaseAuthUI/Sources/FUIAccountSettingsOperationDeleteAccount.m
@@ -43,7 +43,8 @@ NS_ASSUME_NONNULL_BEGIN
                                   alertMessage:FUILocalizedString(kStr_DeleteAccountBody)
                               alertCloseButton:FUILocalizedString(kStr_Cancel)
                                providerHandler:^(id<FIRUserInfo> provider) {
-    if (![provider.providerID isEqualToString:FIREmailAuthProviderID]) {
+    // TODO: Use public API after Firebase 11 minimum.
+    if (![provider.providerID isEqualToString:@"email"]) {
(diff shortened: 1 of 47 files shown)


FirebaseAuthUI/Sources/FUIAccountSettingsOperationDeleteAccount.m
```diff
@@ -43,7 +43,8 @@ - (void)showDeleteAccountDialog {
                                   alertMessage:FUILocalizedString(kStr_DeleteAccountBody)
                               alertCloseButton:FUILocalizedString(kStr_Cancel)
                                providerHandler:^(id<FIRUserInfo> provider) {
-    if (![provider.providerID isEqualToString:FIREmailAuthProviderID]) {
+    // TODO: Use public API after Firebase 11 minimum.
+    if (![provider.providerID isEqualToString:@"email"]) {
```
morganchen12, line 47: optional: Since this is across multiple files can you define a temporary module-level constant and put the TODO over that constant? Applies to the other strings as well.

Optional since these can all be found with code search.

morganchen12 (approved):

FirebaseAuthUI/Sources/FUIAccountSettingsOperationDeleteAccount.m
paulb777: Going to skip this since it would require a separate header for each product or more rearchitecture of both build systems.

PR #1237, opened 2025-02-18
feat: init SwiftUI library for FirebaseAuthSwiftUI

russellwheatley:
**Work in progress**


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FirebaseAuthSwiftUI.swift
```diff
@@ -0,0 +1,102 @@
...
+protocol AuthListenerProtocol {
+    func onError(_ error: Error)
+    func onBeforeSignIn()
+    func onSignedIn(_ user: User)
+    func onCanceled()
+    // TODO - add when I get to this point
+     func onCredentialReceived(_ credential: AuthCredential)
+     func onCredentialLinked(_ credential: AuthCredential)
+    // func onMFARequired(_ resolver: MultiFactorResolver)
+}
```
morganchen12, line 17: Since we're targeting this toward SwiftUI users, what do you think of adding a `@StateObject` that tracks the auth state so consumers can use it directly instead of having to set up their own wrapper? This could be the auth provider itself or a separate object that's managed by the auth provider, or a default implementation of the `AuthListenerProtocol`.

morganchen12 (commented):
Overall looks good. Had one architecture question for consideration.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -0,0 +1,197 @@
...
+  public var body: some View {
+    VStack {
+      title
+      emailAuthButton
+    }.modifier(vStackModifier)
+  }
+
+  // Default implementation that can be overridden
+  public var title: AnyView {
+    AnyView(
```
peterfriese, line 130: Please avoid AnyView. See https://www.swiftbysundell.com/articles/avoiding-anyview-in-swiftui/ and https://youtu.be/YjSxPxT5V40?si=fduquymmiDHmQD86&t=1026.

Annotate the property as `@ViewBuilder` instead. It also probably doesn't need to be `public`, since you're only using it from the `body` of `AuthPickerView`.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  public var body: some View {
+    VStack {
+      AnyView(authPickerView)
+    }
+  }
+}
+
+public protocol AuthPickerViewProtocol: View {
```
peterfriese, line 96: What's the purpose of this protocol?

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -0,0 +1,197 @@
...
+}
+
+public struct AuthPickerView<Modifier: ViewModifier>: AuthPickerViewProtocol {
+  private var emailAuthButton: EmailAuthButton
+  private var vStackModifier: Modifier
+
+  public init(title _: String? = nil, _emailAuthButton: EmailAuthButton? = nil,
+              _modifier: Modifier? = nil) {
+    emailAuthButton = _emailAuthButton ?? EmailAuthButton()
+    vStackModifier = _modifier ?? AuthPickerModifier() as! Modifier
```
peterfriese, line 118: Avoid force unwrapping

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -0,0 +1,197 @@
...
+        buttonContent
+      }
+      NavigationLink(destination: EmailEntryView(), isActive: $emailAuthView) {
+        EmptyView()
+      }
+    }
+  }
+
+  // Default implementation that can be overridden
+  public var buttonContent: AnyView {
```
peterfriese, line 159: Please avoid using `AnyView`.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -0,0 +1,197 @@
...
+    AnyView(
+      Text("Sign in with email")
```
peterfriese, line 161: I recommend using `Button` and (if required) define custom button styles. See https://peterfriese.dev/blog/2023/swiftui-styling-views/

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -0,0 +1,197 @@
...
+  @State private var email: String = ""
+  @EnvironmentObject var authFUI: FUIAuth
+
+  public var body: some View {
+    VStack {
+      Text("Email")
+        .font(.largeTitle)
+        .padding()
+      TextField("Email", text: $email, onCommit: emailSubmit)
+        .textFieldStyle(RoundedBorderTextFieldStyle())
```
peterfriese: Use `.roundedBorder` as a shorthand instead: https://developer.apple.com/documentation/swiftui/roundedbordertextfieldstyle#overview

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -160,38 +169,113 @@ public struct EmailAuthButton: FUIButtonProtocol {
...
+          .font(.body)
+          .padding()
+          .background(Color.blue)
+          .foregroundColor(.white)
+          .cornerRadius(8)
+      }
     }
-    .padding(20)
+    .frame(width: 300, height: 150)
     .background(Color.white)
```
peterfriese, line 235: Make sure to use system colors to ensure this works well in light / dark mode.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FUIAuth.swift
```diff
@@ -49,7 +58,7 @@ public class FUIAuth: ObservableObject {
   }
 
   public func getEmailProvider() -> EmailAuthProvider? {
-    return try! providerWithId(providerId: "email") as! EmailAuthProvider
+    return try! providerWithId(providerId: "password") as! EmailAuthProvider
```
peterfriese, line 61: Please avoid using force unwrap.

Would also recommend using constants (enum, probably) for the provider names.

peterfriese (changes requested):

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/FirebaseAuthSwiftUI.swift
russellwheatley: Hey @morganchen12 - sorry for the lates response, I was on holiday last week, and this week I wanted something more concrete to show you so we can agree on top level API design/flow. I still need to refactor a couple of things, but I'm hoping by the end of the day tomorrow I can provide how it works and see if you think it is reasonable. Also - agree on the point of having some form of a state object the user can consume for auth events 👍

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/Views/FUIAuthPicker.swift
```diff
@@ -28,18 +28,23 @@ public protocol AuthPickerViewProtocol: View {
   var configuration: AuthPickerViewConfiguration { get }
 }
 
-public struct AuthPickerView: AuthPickerViewProtocol {
+public struct AuthPickerView<Content: View>: AuthPickerViewProtocol {
   public let configuration: AuthPickerViewConfiguration
+  private let content: Content
 
-  public init(configuration: AuthPickerViewConfiguration = AuthPickerViewConfiguration()) {
+  public init(configuration: AuthPickerViewConfiguration = AuthPickerViewConfiguration(),
```
peterfriese, line 35: The configuration would probably be either a view modifier, or better yet, a style.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/FirebaseAuthSwiftUI/Views/Warning.swift
```diff
@@ -0,0 +1,80 @@
...
+  public var buttonCornerRadius: CGFloat
+  public var viewBackgroundColor: Color
+  public var viewCornerRadius: CGFloat
+  public var shadowRadius: CGFloat
+  public var strokeColor: Color
+  public var strokeLineWidth: CGFloat
+  public var frameWidth: CGFloat
+  public var frameHeight: CGFloat
+
+  public init(warningMessage: String = "Incorrect email address",
```
peterfriese, line 19: All of these should be view modifiers.

FirebaseSwiftUI/FirebaseEmailAuthSwiftUI/Sources/FirebaseEmailAuthSwiftUI/Views/EmailAuthButton.swift
```diff
@@ -0,0 +1,51 @@
+import SwiftUI
+
+public protocol FUIButtonProtocol: View {
+  var configuration: EmailAuthButtonConfiguration { get }
+}
+
+public class EmailAuthButtonConfiguration {
+  public var buttonText: String
```
peterfriese, line 8: In SwiftUI, you configure views by using view modifiers. They allow for a much more scalable configuration of views.

FirebaseSwiftUI/FirebaseEmailAuthSwiftUI/Sources/FirebaseEmailAuthSwiftUI/Views/EmailEntry.swift
```diff
@@ -0,0 +1,45 @@
+import SwiftUI
+
+public struct EmailEntryView: View {
+  @State private var email: String = ""
+  @State private var invalidEmailWarning: Bool = false
+  @EnvironmentObject var authFUI: FUIAuth
+
+  public var body: some View {
+    if invalidEmailWarning {
```
peterfriese, line 9: This will break view identity - see https://swiftwithmajid.com/2021/12/09/structural-identity-in-swiftui/

russellwheatley:
Hey @morganchen12 & @peterfriese - I made some changes based heavily on what Peter's repo example (much appreciated 🙏 ). Please disregard styling for now until we have some designs (as agreed in previous meeting).

I was wondering about EmailPasswordView whether it is worth separating Sign-In, Login & Password recovery into their own Views and conditionally render based on observable property for AuthenticationFlow? I guess this would allow the user to drop them in if they wish.

I think we also need have some form of initialisation of config for localisation, auth instance & autoAnonymousLogin so any thoughts on that would be appreciated, possibly on the AuthEnvironment observable instance.

FirebaseSwiftUI/FirebaseGoogleSwiftUI/Sources/Services/GoogleProviderSwift.swift
```diff
@@ -0,0 +1,59 @@
+@preconcurrency import FirebaseAuth
+import FirebaseAuthSwiftUI
+import GoogleSignIn
+
+let kGoogleUserInfoEmailScope = "https://www.googleapis.com/auth/userinfo.email"
+let kGoogleUserInfoProfileScope = "https://www.googleapis.com/auth/userinfo.profile"
+let kDefaultScopes = [kGoogleUserInfoEmailScope, kGoogleUserInfoProfileScope]
+
+public enum GoogleProviderError: Error {
+  case rootViewController(String)
```
peterfriese, line 10: Should be rootViewControllerNotFound

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -0,0 +1,343 @@
...
+
+public enum AuthServiceError: Error {
+  case invalidEmailLink(String)
+  case notConfiguredProvider(String)
+  case clientIdNotFound(String)
+  case notConfiguredActionCodeSettings(String)
+}
+
+@MainActor
+final class AuthListenerManager {
```
peterfriese, line 36: Make this private.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/StringUtils.swift
```diff
@@ -0,0 +1,58 @@
...
+let kEmailAlreadyInUseError = "EmailAlreadyInUseError"
+let kInvalidEmailError = "InvalidEmailError"
+let kWeakPasswordError = "WeakPasswordError"
+let kSignUpTooManyTimesError = "SignUpTooManyTimesError"
+let kWrongPasswordError = "WrongPasswordError"
+let kAccountDisabledError = "AccountDisabledError"
+let kEmailsDoNotMatchError = "EmailsDoNotMatchError"
+let kUnknownError = "UnknownError"
+
+public class StringUtils {
```
peterfriese, line 16: Let's see if using SwiftUI's string localization is a more canonical way to support this.

See https://developer.apple.com/documentation/xcode/localizing-and-varying-text-with-a-string-catalog

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -0,0 +1,343 @@
...
+    } catch {
+      throw error
+    }
+  }
+
+  func handleSignInLink(url url: URL) async throws {
+    do {
+      guard let email = emailLink else {
+        throw AuthServiceError.invalidEmailLink(
+          "Invalid email address. Most likely, the link you used has expired. Try signing in again."
```
morganchen12, line 269: Should this say `Invalid sign in link` instead of `Invalid email address`?

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Utils/FacebookUtils.swift
```diff
@@ -0,0 +1,42 @@
+import CommonCrypto
+import Foundation
+import Security
+
+class FacebookUtils {
+  static func randomNonce(length: Int = 32) -> String {
```
peterfriese, line 6: This looks very similar to the nonce code we need to implement for Apple.

morganchen12 (changes requested):
Everything that's marked TODO can be done in a future PR.

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Views/FacebookButtonView.swift
```diff
@@ -0,0 +1,108 @@
...
+public struct FacebookButtonView {
+  @Environment(AuthService.self) private var authService
+  @State private var errorMessage = ""
+  @State private var showCanceledAlert = false
+  @State private var showUserTrackingAlert = false
+  @State private var limitedLogin = true
+
+  public init() {}
+
+  private var limitedLoginBinding: Binding<Bool> {
```
peterfriese, line 18: This computed property does a lot. It is trying to sync with ATTrackingManager.trackingAuthorizationStatus, and also drives showUserTrackingAlert and limitedLogin.

We should simplify this.

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Views/FacebookButtonView.swift
```diff
@@ -0,0 +1,108 @@
...
+    Binding(
+      get: { self.limitedLogin },
+      set: { newValue in
+        let trackingStatus = ATTrackingManager.trackingAuthorizationStatus
+        if newValue == true, trackingStatus != .authorized {
+          self.showUserTrackingAlert = true
```
peterfriese, line 24: Instead of changing this from inside a computed property, I would recommend using the `onChange(of:)` view modifier to listen to `limitedLogin` and update `showUserTrackingAlert` accordingly.

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Views/FacebookButtonView.swift
peterfriese: Here is an alternative solution that feels cleaner:

```
  @State private var showUserTrackingAlert = false
  @State private var trackingAuthorizationStatus: ATTrackingManager.AuthorizationStatus = .notDetermined

  var isLimitedLogin: Bool {
    trackingAuthorizationStatus != .authorized
  }

  public init() {
    _trackingAuthorizationStatus = State(initialValue: ATTrackingManager.trackingAuthorizationStatus)
  }

  func requestTrackingPermission() {
    ATTrackingManager.requestTrackingAuthorization { status in
      Task { @MainActor in
        trackingAuthorizationStatus = status
        if status != .authorized {
          showUserTrackingAlert = true
        }
      }
    }
  }
```

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -0,0 +1,343 @@
...
+  func reauthenticate() async throws {
+    if let user = auth.currentUser, let credential = signedInCredential {
+      try await user.reauthenticate(with: credential)
+    }
+  }
+
+  func deleteUser() async throws {
+    do {
+      if let user = auth.currentUser, let lastSignInDate = user.metadata.lastSignInDate {
+        let needsReauth = !lastSignInDate.isWithinPast(minutes: 5)
```
peterfriese, line 207: Instead of making assumptions about the grace period, it'd be better to make the call and handle the error.

See https://github.com/peterfriese/IdentityKit/blob/a9383206ef3dce073e3b64b6828ccb8d3b77315b/Sources/IdentityKit/Service/AccountService.swift#L68

peterfriese (changes requested):

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Views/FacebookButtonView.swift
russellwheatley: updated to reduce scope of binding: [`0d5d8ba` (#1237)](https://github.com/firebase/FirebaseUI-iOS/pull/1237/commits/0d5d8baa0dc001c5c72203e6c184d90e9400a215)

Not sure if you agree but a binding was still needed to stop toggle from switching if limited login wasn't possible.

samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExample/FirebaseSwiftUIExampleApp.swift
```diff
@@ -0,0 +1,110 @@
...
+  var body: some Scene {
+    WindowGroup {
+      NavigationView {
+        ContentView()
+      }
+    }
+  }
+}
+
+struct ContentView: View {
```
peterfriese, line 77: TODO: please move the `ContentView` into a separate file to make it easier to find.

FirebaseSwiftUI/FirebasePhoneAuthSwiftUI/Sources/Views/PhoneAuthButtonView.swift
```diff
@@ -0,0 +1,92 @@
...
+          TextField("Enter phone number", text: $phoneNumber)
+            .textInputAutocapitalization(.never)
+            .disableAutocorrection(true)
+            .submitLabel(.next)
+        } label: {
+          Image(systemName: "at")
+        }.padding(.vertical, 6)
+          .background(Divider(), alignment: .bottom)
+          .padding(.bottom, 4)
+        Button(action: {
```
peterfriese, line 30: TODO: 
Nit: I would prefer the following signature (here, and in general):

```swift
    Button {
    } label: {
    }
```

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
```diff
@@ -0,0 +1,51 @@
...
+      if authService.authenticationState == .authenticated {
+        SignedInView()
+      } else if authService.authView == .passwordRecovery {
+        PasswordRecoveryView()
+      } else if authService.authView == .emailLink {
+        EmailLinkView()
+      } else {
+        Text(authService.authenticationFlow == .login ? "Login" : "Sign up")
+        VStack { Divider() }
+        EmailAuthView()
```
peterfriese, line 30: So this means that we'll always have Email/Password be the most prominent provider. I would advocate to allow users to instantiate EmailAuthView themselves via the providerButtons closure.

1) I think we should no longer promote Email/Password as the go-to auth provider. Passwords are inherently insecure, and we should promote more secure providers instead.
2) This gives developers more flexibility (e.g. put SiwA / SiwG first, and Email last)

FirebaseSwiftUI/FirebaseGoogleSwiftUI/Sources/Services/GoogleProviderSwift.swift
```diff
@@ -0,0 +1,61 @@
...
+let kGoogleUserInfoProfileScope = "https://www.googleapis.com/auth/userinfo.profile"
+let kDefaultScopes = [kGoogleUserInfoEmailScope, kGoogleUserInfoProfileScope]
+
+public enum GoogleProviderError: Error {
+  case rootViewControllerNotFound(String)
+  case authenticationToken(String)
+  case user(String)
+}
+
+public class GoogleProviderSwift: @preconcurrency GoogleProviderProtocol {
```
peterfriese, line 15: I'm not super happy with the `Swift` suffix. Also, can we include `Auth` in the name? Same for the protocol(s).

This applies to all providers.

peterfriese (approved):
Looking good. I left a few comments, happy for you do address them in a follow-up PR.

I'd also like to see the following:
1) Previews for all views. At the moment, it's not possible to preview the views, which makes it difficult to judge the design.
2) Package.swift files for the individual modules (to support (1))

Approved provided my remarks are addressed in follow-up PRs.

russellwheatley:
Thanks, @peterfriese. Now the last bits of feedback have been resolved, and as discussed in the last meeting, I will merge this PR. I'll follow up with the further suggestions you've made with the exception of email/password View update. This is on hold until we have finalised solution for using View modifiers ([See this PR](https://github.com/firebase/FirebaseUI-iOS/pull/1243)). But I think we're all in agreement, users should be able to decide if they want email/password auth which will be opted-in (i.e. something like `AuthService().withEmailPasswordSignIn()`).

PR #1243, opened 2025-04-17
Move buttons out of ViewBuilder

morganchen12:


FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Services/FacebookProviderSwift.swift
```diff
@@ -34,6 +35,10 @@ public class FacebookProviderSwift: FacebookProviderProtocol {
     shaNonce = CommonUtils.sha256Hash(of: rawNonce)
   }
 
+  @MainActor public var authButton: any View {
```
russellwheatley, line 38: I'm going to presume we should be avoiding "any View" as [mentioned in the main PR](https://github.com/firebase/FirebaseUI-iOS/pull/1237#discussion_r1983491150), unless it is unavoidable. Thoughts, @peterfriese?

russellwheatley (changes requested):
@morganchen12 - Aside from the question about the use of "Any View", it looks fine to me.

russellwheatley (approved):

russellwheatley:
LGTM - I'll continue the same setup for the other providers.

PR #1244, opened 2025-04-24
refactor: follow up PR with updates based on feedback

russellwheatley:
Contains the following:

1. Refactored the protocol and class names for providers so they contain `AuthUI`. Example - `FacebookProviderSwift` has changed to `FacebookProviderAuthUI`.
2. Updated Button Views with `label` argument rather than using anonymous closure.
3. Implemented `ContentView` for SwiftUI example app.
4. Created Previews for almost all Views, needed a FirebaseApp configured for a lot as we initialise FirebaseAuth in AuthService. I didn't create a preview for AuthPicker as it will be refactored soon with the implementation of View modifiers to choose authentication methods available. I will add it once that has been refactored.

Let me know what you think, @peterfriese. Just as an aside, the UI Views don't look great at the moment as we're awaiting UI design.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/CommonUtils.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/CommonUtils.swift
@@ -47,8 +47,10 @@ public class CommonUtils {
     }
     return hash.map { String(format: "%02x", $0) }.joined()
   }
+}
 
-  public static func dummyConfigurationForPreview() {
+public extension FirebaseOptions {
+  static func dummyConfigurationForPreview() {
     guard FirebaseApp.app() == nil else { return }
 
     let options = FirebaseOptions(
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/CommonUtils.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/CommonUtils.swift
@@ -1,4 +1,5 @@
 import CommonCrypto
+import FirebaseCore
 import Foundation
 import Security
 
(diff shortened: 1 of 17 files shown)


samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExample/ContentView.swift
```diff
@@ -0,0 +1,48 @@
...
+import FirebaseFacebookSwiftUI
+import FirebaseGoogleSwiftUI
+import FirebasePhoneAuthSwiftUI
+import SwiftUI
+
+struct ContentView: View {
+  let authService: AuthService
+
+  init() {
+    // Auth.auth().signInAnonymously()
```
peterfriese, line 19: Are you planning to uncomment this?

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
```diff
@@ -38,11 +38,11 @@ extension AuthPickerView: View {
             withAnimation {
               switchFlow()
             }
-          }) {
+          }, label: {
```
peterfriese, line 41: Sorry, I think I might have been a bit vague when asking for this change.

I think I'd prefer the following syntax:

```
            Button(action: { presentingFilters.toggle() }) {
              Text("Unread")
            }
```

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/CommonUtils.swift
```diff
@@ -46,4 +47,19 @@ public class CommonUtils {
     }
     return hash.map { String(format: "%02x", $0) }.joined()
   }
+
+  public static func dummyConfigurationForPreview() {
```
peterfriese, line 51: I wonder if we should make this an extension on `FirebaseOptions` instead.

peterfriese (changes requested):
Left a few more comments

peterfriese (approved):

PR #1247, opened 2025-04-29
feat: anonymous account upgrade with error handling

russellwheatley:
## Notes
1. Set up auto upgrade anonymous account for:
- Signing user in with credential
- Creating user
- Sign in with email link

2. Created merge conflict error so we can pass back the credential to the user should it fail on `email-already-in-use` like in the current FirebaseUI Auth. This is logic is duplicated in a few place in previous implementation but is handled in one spot for FUIAuth SwiftUI.

## Questions
1. Do we want the logic for forcing same device? It isn't a configuration option currently in FUIAuth SwiftUI. See: https://github.com/firebase/FirebaseUI-iOS/blob/main/FirebaseEmailAuthUI/Sources/FUIEmailAuth.m#L829-L844. Personally, I think this might be overkill. At the moment, if `email-link` isn't in app storage, we throw an error here: https://github.com/firebase/FirebaseUI-iOS/pull/1247/files#diff-659d4c231d40710232200eb4482235de5964068a1395944bc32983fe0ed8670cR305. We could just push user back to email link sign-in View if it isn't present.
2. Do we want to also link provider with email link? In the previous implementation, it does have this ability but only for Facebook as far as I can tell. It checks query param here: https://github.com/firebase/FirebaseUI-iOS/blob/main/FirebaseEmailAuthUI/Sources/FUIEmailAuth.m#L261-L270 but only for Facebook: https://github.com/firebase/FirebaseUI-iOS/blob/main/FirebaseEmailAuthUI/Sources/FUIEmailAuth.m#L296-L359


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -169,19 +169,19 @@ public final class AuthService {
 
   public func signIn(credentials credentials: AuthCredential) async throws {
     authenticationState = .authenticating
-    if currentUser?.isAnonymous == true, configuration.shouldAutoUpgradeAnonymousUsers {
-      try await linkAccounts(credentials: credentials)
-    } else {
-      do {
+    do {
+      if currentUser?.isAnonymous == true, configuration.shouldAutoUpgradeAnonymousUsers {
```
russellwheatley, line 173: Question - If the user is signed-in anonymously,  and the user subsequently signs in with email credential, it will raise the exception `email-already-in-use`. If we get this exception, do we just sign the user in with email credential without linking?

russellwheatley: Also - there should be an opportunity to merge backend data if linking fails, some form of callback to be run before signing user in with credential. This code snippet demonstrates what I mean (I couldn't find equivalent iOS sample): https://github.com/firebase/snippets-web/blob/467eaa165dcbd9b3ab15711e76fa52237ba37f8b/snippets/auth-next/link-multiple-accounts/auth_merge_accounts.js#L8

morganchen12: You can copy the merge behavior in the ObjC FirebaseUI implementation here: https://github.com/firebase/FirebaseUI-iOS/blob/main/FirebaseEmailAuthUI/Sources/FUIEmailAuth.m#L397

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -167,21 +173,37 @@ public final class AuthService {
     }
   }
 
+  public func handleAutoUpgradeAnonymousUser(credentials credentials: AuthCredential) async throws {
+    do {
+      try await currentUser?.link(with: credentials)
```
morganchen12, line 178: We should handle the `currentUser` null state in a separate conditional, otherwise this do-catch has the possibility of executing nothing and returning no error. From a convenience perspective it may also be nice to pass the unwrapped user through to the `AccoutMergeConflictContext` so the consumer doesn't have to again unwrap `currentUser`.

In theory `currentUser` should never be null here, so throwing an error is appropriate.

morganchen12 (commented):
LGTM other than the one comment.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -167,21 +173,37 @@ public final class AuthService {
     }
   }
 
+  public func handleAutoUpgradeAnonymousUser(credentials credentials: AuthCredential) async throws {
```
peterfriese, line 176: No need to duplicate the argument label if you don't want to use a different name inside the function. This can just be

`public func handleAutoUpgradeAnonymousUser(credentials: AuthCredential) async throws {`

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -167,21 +173,37 @@ public final class AuthService {
...
+          underlyingError: error,
+          message: "Unable to merge accounts. Use the credential in the context to resolve the conflict."
+        )
+        throw AuthServiceError.accountMergeConflict(context: context)
+      }
+      throw error
+    }
+  }
+
   public func signIn(credentials credentials: AuthCredential) async throws {
```
peterfriese, line 192: Remove duplicate argument label.

`public func signIn(credentials: AuthCredential) async throws {`

peterfriese (changes requested):

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
russellwheatley: @morganchen12 - updated. I didn't add User to the exception as it produced an Xcode compiler error: Stored property 'user' of 'Sendable'-conforming struct 'AccountMergeConflictContext' has non-sendable type 'User'

It seems User is non-sendable.

morganchen12: A fix for that is in the pipeline in upstream Firebase, so maybe leave a TODO and just send the user's UID (String) for now?

russellwheatley: Done, see [`a014c7a` (#1247)](https://github.com/firebase/FirebaseUI-iOS/pull/1247/commits/a014c7a548f11031ee086877c1e37f9e43790d2d)

PR #1248, opened 2025-04-30
feat: utilise string localisations

russellwheatley:
PR contains string localization implementation.

A couple of things to note;

- There are no strings as far as I can see for verifying email so I left VerifyEmailView as it is, until the below question is resolved.
- I haven't touched the Facebook, Phone or Google provider Swift package until the question below has been resolved.

**Questions**
1. Should we keep all strings within the core package?  If they need to be within their own Swift package (e.g. Facebook strings in Facebook Swift package), then this would require some refactoring to the internal implementation so we can make use of each packages `Bundle.module` which contains the strings.
2. Should we add strings to the string catalog for missing strings such as verifying email?

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Strings/Localizable.xcstrings
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Strings/Localizable.xcstrings
@@ -394,13 +394,25 @@
       }
     },
     "EmailLinkSignInLabel" : {
-      "comment" : "Button label for email link sign-in",
+      "comment" : "Button label to push user to email link sign-in",
       "extractionState" : "manual",
       "localizations" : {
         "en" : {
           "stringUnit" : {
             "state" : "translated",
             "value" : "Prefer Email link sign-in?"
+          }
+        }
+      }
+    },
+    "EmailLinkSignInTitle" : {
+      "comment" : "Sign in with email link View title",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Sign in with email link"
           }
         }
       }
@@ -542,9 +554,6 @@
           }
         }
       }
-    },
-    "Key" : {
-      "extractionState" : "manual"
     },
     "Log in" : {
 
@@ -842,6 +851,18 @@
     },
     "Send email sign-in link" : {
 
+    },
+    "SendEmailSignInLinkButtonLabel" : {
+      "comment" : "Button label for sending email sign-in link",
+      "extractionState" : "manual",
+      "localizations" : {
+        "en" : {
+          "stringUnit" : {
+            "state" : "translated",
+            "value" : "Send email sign-in link"
+          }
+        }
+      }
     },
     "Sign in with email link" : {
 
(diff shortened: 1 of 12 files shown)


morganchen12:
> Should we want to keep all strings within the core package? If they need to be within their own Swift package (e.g. Facebook strings in Facebook Swift package), then this would require some refactoring to the internal implementation so we can make use of each packages Bundle.module which contains the strings.

Though it's a big of an abstraction-break I think we should given the localization bundle override is in the core package and strings themselves aren't stateful and have no dependencies. Overall I expect the strings required for 3P auth providers to be minimal since those dependencies (GoogleSignIn etc) also do their own localization.

> Should we add strings to the string catalog for missing strings such as verifying email?

Yes, if you can add placeholders that'd be great. I'll fill them in with internal l10n resources async and they shouldn't be a blocker for this project. In the meantime Xcode will display whatever you use as a placeholder.

morganchen12 (approved):

peterfriese (changes requested):
Is there a reason why we're not using SwiftUI's built-in string localisation?

See:
* [Localize your SwiftUI app - WWDC21 - Videos - Apple Developer](https://developer.apple.com/videos/play/wwdc2021/10220)
* [Preparing views for localization | Apple Developer Documentation](http://go/appledoc/swiftui/preparing-views-for-localization)
* [Localizing and varying text with a string catalog | Apple Developer Documentation](http://go/appledoc/xcode/localizing-and-varying-text-with-a-string-catalog)

russellwheatley:
@peterfriese -  The reason why I kept computed properties instead of using string literals (I think this is what you mean by built-in string localization) is for two reasons: 

1. It allows us to document the strings inline, and makes it pretty clear which strings are localised.
2. It allows us to control the bundle passed into configuration used for localized strings: https://github.com/firebase/FirebaseUI-iOS/blob/string-localisation/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/StringUtils.swift#L19-L27

Happy to update if you prefer an alternative.

peterfriese:
> @peterfriese - The reason why I kept computed properties instead of using string literals (I think this is what you mean by built-in string localization) is for two reasons:
> 
> 1. It allows us to document the strings inline, and makes it pretty clear which strings are localised.

This should be covered by the `comment` parameter (see https://developer.apple.com/documentation/swiftui/preparing-views-for-localization#Add-comments-to-text-views):

```
Text("Explore",
     comment: "The title of the tab bar item that navigates to the Explore screen.")
```

> 2. It allows us to control the bundle passed into configuration used for localized strings: https://github.com/firebase/FirebaseUI-iOS/blob/string-localisation/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Utils/StringUtils.swift#L19-L27
> 

In which situations would this be used?

russellwheatley:
@peterfriese - There is a `customStringsBundle` in configuration that allows the user to customize strings which is specified in the design doc.

I'm not sure how we control the bundle if we use string literals. Also -  we have to specify the `module` bundle like so to retrieve the value from the string catalog:

```swift
Text("VerifyItsYou", bundle: Bundle.module)
```

If we do this:

```swift
Text("VerifyItsYou")
```

It'll render `VerifyItsYou` as a fallback presumably because it is looking in the app bundle for the key/value.

peterfriese:
> If we do this:
> 
> ```swift
> Text("VerifyItsYou")
> ```
> 
> It'll render `VerifyItsYou` as a fallback presumably because it is looking in the app bundle for the key/value.

You would use the actual English phrase in the UI, so you wouldn't run into this issue. See https://developer.apple.com/documentation/swiftui/preparing-views-for-localization and https://developer.apple.com/documentation/Xcode/localizing-and-varying-text-with-a-string-catalog

russellwheatley:
@peterfriese - Yes, you would get the English translation but you wouldn't get any other translation if we didn't specify `Bundle.module`.

PR #1249, opened 2025-05-01
refactor: the way providers are initialized. `AnyView` is used to render provider buttons.

russellwheatley:
I cannot find a way to make it work where the Buttons are not type erased. 

We have types unknown to core auth package and we're trying to render unknown button types in core auth package. How can we render without using type erasure?

If there is a clever way of doing this, I would love to know.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
@@ -6,15 +6,15 @@ public protocol ExternalAuthProvider {
   @MainActor func authButton() -> AnyView
 }
 
-public protocol GoogleProviderProtocol: ExternalAuthProvider {
+public protocol GoogleProviderAuthUIProtocol: ExternalAuthProvider {
   @MainActor func signInWithGoogle(clientID: String) async throws -> AuthCredential
 }
 
-public protocol FacebookProviderProtocol: ExternalAuthProvider {
+public protocol FacebookProviderAuthUIProtocol: ExternalAuthProvider {
   @MainActor func signInWithFacebook(isLimitedLogin: Bool) async throws -> AuthCredential
 }
 
-public protocol PhoneAuthProviderProtocol: ExternalAuthProvider {
+public protocol PhoneAuthProviderAuthUIProtocol: ExternalAuthProvider {
   @MainActor func verifyPhoneNumber(phoneNumber: String) async throws -> String
 }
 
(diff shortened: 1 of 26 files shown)


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -2,8 +2,8 @@
 import SwiftUI
 
 public protocol ExternalAuthProvider {
-  associatedtype ButtonType: View
-  @MainActor var authButton: ButtonType { get }
+  var id: String { get } 
+  @MainActor func authButton() -> AnyView
```
peterfriese, line 6: Did you try returning `some View`? You might need to annotate the function with `@ViewBuilder`

russellwheatley: cannot have `some View` as return type on protocol apparently. Get Xcode  compiler error: `'some' type cannot be the return type of a protocol requirement; did you mean to add an associated type?`

peterfriese: I put together something that might help: https://gist.github.com/peterfriese/d9745f366fb4f857daf077e93b1eb01f

```
import SwiftUI

protocol AuthProvider: Identifiable {
  var id: String { get }
  associatedtype ButtonType: View
  var authButtonView: Self.ButtonType { get }
}

protocol FooProviderProtocol: AuthProvider {
  func doFoo()
}

struct FooProviderImpl1: FooProviderProtocol {
  var id: String = "FooProviderImpl1"

  func doFoo() {
    print("1")
  }

  var authButtonView: some View {
    Text("1")
  }
}

struct FooProviderImpl2: FooProviderProtocol {
  var id: String = "FooProviderImpl2"

  func doFoo() {
    print("2")
  }

  var authButtonView: some View {
    Text("2")
      .background(Color.red)
  }
}

protocol BarProviderProtocol: AuthProvider {
  func doBar()
}

struct BarProviderImpl1: BarProviderProtocol {
  var id: String = "BarProviderImpl1"
  
  func doBar() {
    print("1")
  }
  
  var authButtonView: some View {
    Text("1")
  }
}

struct BarProviderImpl2: BarProviderProtocol {
  var id: String = "BarProviderImpl2"
  
  func doBar() {
    print("2")
  }
  
  var authButtonView: some View {
    Text("2")
      .background(Color.red)
  }
}


struct ContentView<Provider: AuthProvider>: View {
  var provider: Provider
  var body: some View {
    VStack {
      provider.authButtonView
    }
  }
}

#Preview {
  ContentView(provider: BarProviderImpl2())
}
```

russellwheatley: I've responded in the other PR: https://github.com/firebase/FirebaseUI-iOS/pull/1250

PR #1250, opened 2025-05-02
example of ViewBuilder issue

russellwheatley:
Hacked together for purpose of demonstration only.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
@@ -2,7 +2,9 @@
 import SwiftUI
 
 public protocol ExternalAuthProvider {
-  var id: String { get } 
+  var id: String { get }
+  associatedtype ButtonType: View
+  @MainActor @ViewBuilder func authButtonView() -> ButtonType
   @MainActor func authButton() -> AnyView
 }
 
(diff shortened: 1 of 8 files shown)


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -96,6 +92,40 @@ public final class AuthService {
...
+        }
+      }
+    )
+  }
+
+  @ViewBuilder
+  public var googleButton: some View {
+    if googleProvider != nil {
+      // For purpose of demonstration
+      // This produces "Type 'any View' cannot conform to 'View'"
```
russellwheatley, line 114: @peterfriese - the problem is when we try to render from core package, Xcode produces this compiler error. I don't know how to get around it.

peterfriese: Can you take a look at https://gist.github.com/peterfriese/d9745f366fb4f857daf077e93b1eb01f and see if this approach helps?

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
```diff
@@ -28,7 +28,8 @@ extension AuthPickerView: View {
         Text(authService.authenticationFlow == .login ? "Login" : "Sign up")
         VStack { Divider() }
         EmailAuthView()
-
+        // Xcode compiler exception: Type 'any ExternalAuthProvider' cannot conform to 'ExternalAuthProvider'
```
russellwheatley, line 31: @peterfriese - Following your gist example, we get this Xcode compiler error now. 

I tried to remove `any ExternalAuthProvider` from the types in auth service for each provider but it isn't allowed because ExternalAuthProvider uses an associated type (ButtonType), Swift requires you to explicitly acknowledge it as an existential by using `any ExternalAuthProvider`.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
russellwheatley: @peterfriese - here is the commit attempting to follow your example: [`8c38462` (#1250)](https://github.com/firebase/FirebaseUI-iOS/pull/1250/commits/8c3846276c21769bbe19530eced2824f8f37f97c)

PR #1252, opened 2025-05-09
refactor: email sign-in opt-in, remove providerButton closure

russellwheatley:
- Allow users to opt-in for email provider sign-in.
- Removed the providerButton closure on AuthPickerView.

## Question
- If the user doesn't opt-in for email provider sign-in, it will also hide email link sign-in. Presumably that is what we want?


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
```diff
@@ -29,28 +26,32 @@ extension AuthPickerView: View {
...
         EmailLinkView()
       } else {
-        Text(authService.authenticationFlow == .login ? authService.string
-          .emailLoginFlowLabel : authService.string.emailSignUpFlowLabel)
-        VStack { Divider() }
-        EmailAuthView()
+        if authService.emailSignInEnabled {
+          Text(authService.authenticationFlow == .login ? authService.string
+            .emailLoginFlowLabel : authService.string.emailSignUpFlowLabel)
+          VStack { Divider() }
```
morganchen12, line 32: This is kind of lazy of me to ask, but why does the divider need to be in another VStack?

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/EmailAuthView.swift
```diff
@@ -127,7 +127,7 @@ extension EmailAuthView: View {
       .frame(maxWidth: .infinity)
       .buttonStyle(.borderedProminent)
       Button(action: {
-        authService.authView = .passwordRecovery
+        authService.authView = .emailLink
```
morganchen12, line 130: Should these be two separate buttons?

russellwheatley: They are two separate buttons because they are two separate Views. 

password recovery: https://github.com/firebase/FirebaseUI-iOS/blob/opt-in-email-provider/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/EmailAuthView.swift#L84

email link sign-in: https://github.com/firebase/FirebaseUI-iOS/blob/opt-in-email-provider/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/EmailAuthView.swift#L129

Which are rendered in the AuthPickerView here: https://github.com/firebase/FirebaseUI-iOS/blob/opt-in-email-provider/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift#L24-L27

If you prefer a different way then let me know 👍

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
russellwheatley: That's a good catch - I got this from the example app Peter provided a while ago. I tested using Preview here: [`d13aea6` (#1252)](https://github.com/firebase/FirebaseUI-iOS/pull/1252/commits/d13aea63bf1e2c0c80af61820969f34a90b28fca)

and found that it made no discernible difference to the UI so I've removed them.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/EmailAuthView.swift
russellwheatley: @morganchen12  Also - I have a question in the description regarding whether calling `AuthService()
    .withEmailSignIn()` should also hide email sign-in. I'm 99% sure that it should but just wanted confirmation.

morganchen12: Do you mean the opposite, where not calling `.withEmailSignIn()` should hide it? That sounds good to me.

russellwheatley: sorry, I did mean the opposite. Thanks for clarification 🙏

morganchen12 (approved):
LGTM

PR #1254, opened 2025-05-12
feat: utilise `tosURL` and `privacyPolicyUrl`. make nullable

russellwheatley:
- made `tosURL` and `privacyPolicyUrl` nullable as discussed in previous meeting, up to user to define.
- Followed existing logic in UIKit FirebaseUI: https://github.com/firebase/FirebaseUI-iOS/blob/main/FirebaseAuthUI/Sources/FUIPrivacyAndTermsOfServiceView.m#L25
- The AuthPickerView is getting a little crowded now, and I imagine it'll have to be broken up into separate navigable Views.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthConfiguration.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthConfiguration.swift
@@ -6,8 +6,8 @@ public struct AuthConfiguration {
   let interactiveDismissEnabled: Bool
   let shouldAutoUpgradeAnonymousUsers: Bool
   let customStringsBundle: Bundle?
-  let tosUrl: URL
-  let privacyPolicyUrl: URL
+  let tosUrl: URL?
+  let privacyPolicyUrl: URL?
   let emailLinkSignInActionCodeSettings: ActionCodeSettings?
   let verifyEmailActionCodeSettings: ActionCodeSettings?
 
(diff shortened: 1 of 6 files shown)


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthConfiguration.swift
```diff
@@ -6,17 +6,17 @@ public struct AuthConfiguration {
   let interactiveDismissEnabled: Bool
   let shouldAutoUpgradeAnonymousUsers: Bool
   let customStringsBundle: Bundle?
-  let tosUrl: URL
-  let privacyPolicyUrl: URL
+  let tosUrl: URL?
+  let privacyPolicyUrl: URL?
```
morganchen12, line 10: nit: with the behavior in the view rendering an empty view if one but not both are set, we should document that both these properties need to be set in order to render a privacy and terms footer.

morganchen12 (approved):

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthConfiguration.swift
russellwheatley: Good catch, I've updated README here: [`37903bd` (#1254)](https://github.com/firebase/FirebaseUI-iOS/pull/1254/commits/37903bded3f61a0923836911754905b079207a6a)

PR #1255, opened 2025-05-14
ui: wrap in ScollView + dividers on AuthPickerView

russellwheatley:

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
@@ -17,45 +17,49 @@ public struct AuthPickerView<Content: View> {
 
 extension AuthPickerView: View {
   public var body: some View {
-    VStack {
-      Text(authService.string.authPickerTitle)
-        .font(.largeTitle)
-        .fontWeight(.bold)
-        .padding()
-      if authService.authenticationState == .authenticated {
-        SignedInView()
-      } else if authService.authView == .passwordRecovery {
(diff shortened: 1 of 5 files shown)


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
```diff
@@ -17,45 +17,51 @@ public struct AuthPickerView<Content: View> {
...
+          .padding()
+        if authService.authenticationState == .authenticated {
+          SignedInView()
+        } else if authService.authView == .passwordRecovery {
+          PasswordRecoveryView()
+        } else if authService.authView == .emailLink {
+          Divider()
+          EmailLinkView()
+        } else if authService.authView == .phoneAuth {
+          // TODO: - how are we rendering the phone auth View??
```
russellwheatley, line 34: This is the nub of the problem and it is the same problem we have rendering the buttons. Core auth package doesn't know about PhoneAuthView so how can we render it?

We could:
1. Use AnyView again like with rendering buttons.
2. Move Phone Auth package in to Core Auth package - A couple hundred lines of code extra, not a tremendous amount.
3. An alternative solution???

russellwheatley: I've updated to present a modal, I've had to follow a similar pattern to `renderingButtons` logic for the same reason: [`dea04f6` (#1255)](https://github.com/firebase/FirebaseUI-iOS/pull/1255/commits/dea04f66ff22e93cc3053363a22ad9a1bff879ef)

PR #1256, opened 2025-05-15
feat: Facebook auth re-authentication to allow deleting user

russellwheatley:
I've made comments in the PR so it is a bit clearer what is going on.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AccountService+Email.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AccountService+Email.swift
@@ -6,7 +6,7 @@ protocol EmailPasswordOperationReauthentication {
 }
 
 extension EmailPasswordOperationReauthentication {
-  func reauthenticate() async throws -> AuthenticationToken {
+  @MainActor func reauthenticate() async throws -> AuthenticationToken {
     guard let user = Auth.auth().currentUser else {
       throw AuthServiceError.reauthenticationRequired("No user currently signed-in")
     }
@@ -28,6 +28,7 @@ extension EmailPasswordOperationReauthentication {
   }
 }
 
+@MainActor
 class EmailPasswordDeleteUserOperation: AuthenticatedOperation,
   EmailPasswordOperationReauthentication {
   let passwordPrompt: PasswordPromptCoordinator
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AccountService+Email.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AccountService+Email.swift
@@ -6,6 +6,7 @@ protocol EmailPasswordOperationReauthentication {
 }
 
 extension EmailPasswordOperationReauthentication {
+  // TODO: - @MainActor because User is non-sendable. Might change this once User is sendable in firebase-ios-sdk
   @MainActor func reauthenticate() async throws -> AuthenticationToken {
     guard let user = Auth.auth().currentUser else {
       throw AuthServiceError.reauthenticationRequired("No user currently signed-in")
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
@@ -83,9 +83,9 @@ public final class AuthService {
   public var authenticationFlow: AuthenticationFlow = .login
   public var errorMessage = ""
   public let passwordPrompt: PasswordPromptCoordinator = .init()
-  private var googleProvider: (any GoogleProviderAuthUIProtocol)?
-  private var facebookProvider: (any FacebookProviderAuthUIProtocol)?
-  private var phoneAuthProvider: (any PhoneAuthProviderAuthUIProtocol)?
+  private var unsafeGoogleProvider: (any GoogleProviderAuthUIProtocol)?
+  private var unsafeFacebookProvider: (any FacebookProviderAuthUIProtocol)?
+  private var unsafePhoneAuthProvider: (any PhoneAuthProviderAuthUIProtocol)?
 
   private var listenerManager: AuthListenerManager?
   private var signedInCredential: AuthCredential?
@@ -94,13 +94,13 @@ public final class AuthService {
   public func register(provider: ExternalAuthProvider) {
     switch provider {
     case let google as GoogleProviderAuthUIProtocol:
-      googleProvider = google
+      unsafeGoogleProvider = google
       providers.append(provider)
     case let facebook as FacebookProviderAuthUIProtocol:
-      facebookProvider = facebook
+      unsafeFacebookProvider = facebook
       providers.append(provider)
     case let phone as PhoneAuthProviderAuthUIProtocol:
-      phoneAuthProvider = phone
+      unsafePhoneAuthProvider = phone
       providers.append(provider)
     default:
       break
@@ -117,9 +117,9 @@ public final class AuthService {
     )
   }
 
-  private var safeGoogleProvider: any GoogleProviderAuthUIProtocol {
+  private var googleProvider: any GoogleProviderAuthUIProtocol {
     get throws {
-      guard let provider = googleProvider else {
+      guard let provider = unsafeGoogleProvider else {
         throw AuthServiceError
           .notConfiguredProvider("`GoogleProviderSwift` has not been configured")
       }
@@ -127,9 +127,9 @@ public final class AuthService {
     }
   }
 
-  private var safeFacebookProvider: any FacebookProviderAuthUIProtocol {
+  private var facebookProvider: any FacebookProviderAuthUIProtocol {
     get throws {
-      guard let provider = facebookProvider else {
+      guard let provider = unsafeFacebookProvider else {
         throw AuthServiceError
           .notConfiguredProvider("`FacebookProviderAuthUI` has not been configured")
       }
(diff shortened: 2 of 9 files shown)


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -82,17 +83,28 @@ public final class AuthService {
   public var authenticationFlow: AuthenticationFlow = .login
   public var errorMessage = ""
   public let passwordPrompt: PasswordPromptCoordinator = .init()
-
-  public var googleProvider: (any GoogleProviderAuthUIProtocol)?
-  public var facebookProvider: (any FacebookProviderAuthUIProtocol)?
-  public var phoneAuthProvider: (any PhoneAuthProviderAuthUIProtocol)?
+  private var googleProvider: (any GoogleProviderAuthUIProtocol)?
+  private var facebookProvider: (any FacebookProviderAuthUIProtocol)?
+  private var phoneAuthProvider: (any PhoneAuthProviderAuthUIProtocol)?
```
russellwheatley, line 88: These are private now because they shouldn't be accessed, should use `safeGoogleProvider`, etc.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -82,17 +83,28 @@ public final class AuthService {
...
+    case let google as GoogleProviderAuthUIProtocol:
+      googleProvider = google
+      providers.append(provider)
+    case let facebook as FacebookProviderAuthUIProtocol:
+      facebookProvider = facebook
+      providers.append(provider)
+    case let phone as PhoneAuthProviderAuthUIProtocol:
+      phoneAuthProvider = phone
+      providers.append(provider)
+    default:
```
russellwheatley, line 105: This now assigns providers correctly without having to manually set + call `register()`.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -247,9 +260,13 @@ public final class AuthService {
...
     do {
-      if let user = auth.currentUser {
-        let operation = EmailPasswordDeleteUserOperation(passwordPrompt: passwordPrompt)
-        try await operation(on: user)
+      if let user = auth.currentUser, let providerId = signedInCredential?.provider {
+        if providerId == "password" {
+          let operation = EmailPasswordDeleteUserOperation(passwordPrompt: passwordPrompt)
+          try await operation(on: user)
+        } else if providerId == "facebook.com" {
+          try await safeFacebookProvider.deleteUser(user: user)
```
russellwheatley, line 268: A bit tricky deleting user with Facebook re-authentication but I found the best way was to have it on the provider itself which makes sense to me anyway.

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Services/AuthService+Facebook.swift
```diff
@@ -10,8 +10,8 @@ import FirebaseAuthSwiftUI
 public extension AuthService {
   @discardableResult
   func withFacebookSignIn(scopes scopes: [String]? = nil) -> AuthService {
-    facebookProvider = FacebookProviderAuthUI(scopes: scopes)
-    register(provider: facebookProvider!)
+    FacebookProviderAuthUI.configureSharedInstance(scopes: scopes)
+    register(provider: FacebookProviderAuthUI.shared)
```
russellwheatley, line 14: Switched to having a private init() for Facebook provider and making it a singleton which is kept on `shared` as there really should only be one instance of it anyway. I can roll this back if you don't like.

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Services/FacebookProviderAuthUI.swift
```diff
@@ -27,35 +22,52 @@ public class FacebookProviderAuthUI: FacebookProviderAuthUIProtocol {
   let shortName = "Facebook"
   let providerId = "facebook.com"
   private let loginManager = LoginManager()
-  private var rawNonce: String
-  private var shaNonce: String
+  private var rawNonce: String?
+  private var shaNonce: String?
+  // Needed for reauthentication
+  var isLimitedLogin: Bool = true
```
russellwheatley, line 28: Need this to track the initial login for re-authentication.

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Services/FacebookProviderAuthUI.swift
```diff
@@ -74,10 +86,8 @@ public class FacebookProviderAuthUI: FacebookProviderAuthUIProtocol {
         case .cancelled:
           continuation
             .resume(throwing: FacebookProviderError.signInCancelled("User cancelled sign-in"))
-        //          showCanceledAlert = true
         case let .failed(error):
           continuation.resume(throwing: error)
-        //          errorMessage = authService.string.localizedErrorMessage(for: error)
```
russellwheatley, line 80: errorMessage and alert are handled upstream, removed here.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AccountService+Email.swift
```diff
@@ -6,7 +6,7 @@ protocol EmailPasswordOperationReauthentication {
 }
 
 extension EmailPasswordOperationReauthentication {
-  func reauthenticate() async throws -> AuthenticationToken {
+  @MainActor func reauthenticate() async throws -> AuthenticationToken {
```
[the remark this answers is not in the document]
morganchen12: SGTM, you could add TODOs here since User will eventually become Sendable but there's a strong argument for just keeping this API `@MainActor` regardless.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
morganchen12: nit: I'd rather have the private apis be named `unsafeGoogleProvider` and the public ones plainly `googleProvider`

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
```diff
@@ -119,7 +131,7 @@ public final class AuthService {
     get throws {
       guard let provider = facebookProvider else {
         throw AuthServiceError
-          .notConfiguredProvider("`FacebookProviderSwift` has not been configured")
+          .notConfiguredProvider("`FacebookProviderAuthUI` has not been configured")
```
morganchen12, line 134: Should this not just be a `fatalError` given consumers won't have conditional configuration of providers?

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Services/FacebookProviderAuthUI.swift
```diff
@@ -109,7 +119,7 @@ public class FacebookProviderAuthUI: FacebookProviderAuthUIProtocol {
     if let idToken = AuthenticationToken.current {
       let credential = OAuthProvider.credential(withProviderID: providerId,
                                                 idToken: idToken.tokenString,
-                                                rawNonce: rawNonce)
+                                                rawNonce: rawNonce!)
```
morganchen12, line 122: It's not immediately clear from the code that this optional will never be `nil` here, so maybe add a comment explaining why it's safe.

morganchen12 (approved):
I'm ok with the singleton behavior. Had a few comments, otherwise LGTM.

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AccountService+Email.swift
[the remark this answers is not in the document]
russellwheatley: SGTM - done 👍

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
russellwheatley: Yeah, that's a better idea, done: [`a454559` (#1256)](https://github.com/firebase/FirebaseUI-iOS/pull/1256/commits/a454559be9e17f1bb7b01b59798b16adce686bbd)

FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Services/AuthService.swift
russellwheatley: Yeah, that makes sense. Updated: [`6b08073` (#1256)](https://github.com/firebase/FirebaseUI-iOS/pull/1256/commits/6b080730ec42991b88aefec55a1e71e1b4627172)

FirebaseSwiftUI/FirebaseFacebookSwiftUI/Sources/Services/FacebookProviderAuthUI.swift
russellwheatley: I've updated code to guard the rawNonce instead as I think that is better: [`38a1b0c` (#1256)](https://github.com/firebase/FirebaseUI-iOS/pull/1256/commits/38a1b0c6d3627673e4c590501a726ff575583cc2)

PR #1280, opened 2025-10-23
refactor: use error alert modal to display errors

russellwheatley:
## Description
- [x] Centralized error modal system implemented
- [x] Automatic error modal display for authentication errors
- [x] Integration with all authentication flows
- [x] Modal dismissal 

- Modal is on [AuthPickerView](https://github.com/firebase/FirebaseUI-iOS/pull/1280/files#diff-d2ca1e6b30ec7a834a4973dd246f3b842496114d02c9bc588901830acbe1eff2R100-R103) which is the parent of all Views (Except PhoneAuthView which is isolated from AuthPickerView).
- PhoneAuthView is a separate modal and also uses the errorModal locally as opposed to using the error modal in AuthPickerView.
- Removed all error messages appearing in Views.
- Updated Password Recovery View to have a success sheet if successful, but falls back to error modal if fails.
- [Wrote a couple of UI tests](https://github.com/firebase/FirebaseUI-iOS/pull/1280/files) for email and Facebook provider to check the modal is activated on sign-in failure.
- There's a bunch of formatting in this PR as well using swiftformat to keep code formatted.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/MFAEnrolmentView.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/MFAEnrolmentView.swift
@@ -77,18 +77,14 @@ public struct MFAEnrolmentView {
   private func startEnrollment() {
     Task {
       isLoading = true
+      defer { isLoading = false }
 
-      do {
-        let session = try await authService.startMfaEnrollment(
-          type: selectedFactorType,
-          accountName: authService.currentUser?.email,
-          issuer: authService.configuration.mfaIssuer
-        )
-        currentSession = session
-        isLoading = false
-      } catch {
-        isLoading = false
-      }
+      let session = try await authService.startMfaEnrollment(
+        type: selectedFactorType,
+        accountName: authService.currentUser?.email,
+        issuer: authService.configuration.mfaIssuer
+      )
+      currentSession = session
     }
   }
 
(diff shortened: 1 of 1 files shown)


FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/MFAEnrolmentView.swift
```diff
@@ -87,11 +85,10 @@ public struct MFAEnrolmentView {
           issuer: authService.configuration.mfaIssuer
         )
         currentSession = session
+        isLoading = false
       } catch {
-        errorMessage = error.localizedDescription
+        isLoading = false
```
demolaf: defer { isLoading = false } outside of do/catch block if you want isLoading to always reset to false

FirebaseSwiftUI/FirebasePhoneAuthSwiftUI/Sources/Views/PhoneAuthView.swift
```diff
@@ -27,7 +27,7 @@ import SwiftUI
 @MainActor
 public struct PhoneAuthView {
   @Environment(\.dismiss) private var dismiss
-  @State private var errorMessage = ""
+  @State private var currentError: AlertError?
```
demolaf: why do we have currentError here and in authService? Is currentError meant to be view specific or app's state?

russellwheatley: Yeah, PhoneAuthView is initiated outside of AuthPickerView so it needs its own errorAlert and we need to track the currentError.

PR #1285, opened 2025-10-27
refactor: AuthPickerView sheet and Navigation stack

demolaf:
## Proof of Concept
- Allow the user to pass in their app via closure on AuthPickerView:
```swift
AuthPickerView {
  usersApp  // Your app's authenticated/unauthenticated content
}
.environment(authService)

var usersApp: some View {
    NavigationStack {
      VStack {
        if authService.authenticationState == .unauthenticated {
          Text("Not Authenticated")
        } else {
          Text("Authenticated - \(authService.currentUser?.email ?? "")")
          Button {
            authService.isPresented = true  // Reopen the sheet so user can manage account
          } label: {
            Text("Manage Account")
          }
          .buttonStyle(.bordered)
          Button {
            Task {
              try? await authService.signOut()
            }
          } label: {
            Text("Sign Out")
          }
          .buttonStyle(.borderedProminent)
        }
      }
      .navigationTitle("Firebase UI Demo")
    }
```
- The AuthPickerView is now wrapped in a `.sheet()`. This means:
	- The authentication picker appears as a modal sheet over your app.
	- It prevents navigation conflicts between the SDK's navigation and your app's navigation.
	- The developer's app content (`usersApp`) is always visible underneath.
	- Can use NavigationStack internally as the sheet provides its own context outside of user's app.
- Implemented our own Navigator class to manage different Auth Views internally (i.e. doesn't affect user's app). This allows us to **pop to root** from any View and clear the stack. For example, when user signs out, we need to clear the stack so they can't swipe back to authenticated View.

## Developer Notes
- NavigationStack for routing
- Updated UI for EmailAuthView, PasswordRecoveryView, PhoneAuthView (now split between EnterPhoneNumberView and EnterVerificationCodeView), EmailLinkView.
- Updated styling for buttons
- Introduced FirebaseAuthUIComponents package which contains resources, UI components and provider styles.
- shouldHideCancelButton in AuthUIConfiguration shows "x" close button in AuthPickerView toolbar
- interactiveDismissEnabled for AuthPickerView sheet
- Removed isLimitedLogin flag from FacebookAuthProviderSwift, App Tracking Transparency (ATT) drives limited login internally

https://github.com/user-attachments/assets/0228305b-b9ff-4421-a13b-d2a6db31c7e7

### DIFF
--- a/samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExampleUITests/TestUtils.swift
+++ b/samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExampleUITests/TestUtils.swift
@@ -32,7 +32,7 @@ func createEmail() -> String {
 // MARK: - User Creation
 
 /// Helper to create a test user in the emulator via REST API (avoids keychain issues)
-func createTestUser(email: String, password: String = "123456",
+@MainActor func createTestUser(email: String, password: String = "123456",
                     verifyEmail: Bool = false) async throws {
   // Use Firebase Auth emulator REST API directly to avoid keychain access issues in UI tests
   let signUpUrl =
@@ -77,7 +77,7 @@ func createTestUser(email: String, password: String = "123456",
 // MARK: - Email Verification
 
 /// Verifies an email address in the emulator using the OOB code mechanism
-func verifyEmailInEmulator(email: String,
+@MainActor func verifyEmailInEmulator(email: String,
                            idToken: String,
                            projectID: String = "flutterfire-e2e-tests",
                            emulatorHost: String = "127.0.0.1:9099") async throws {


samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExample/App/ContentView.swift
```diff
@@ -0,0 +1,175 @@
...
+      tosUrl: URL(string: "https://example.com/tos"),
+      privacyPolicyUrl: URL(string: "https://example.com/privacy"),
+      emailLinkSignInActionCodeSettings: actionCodeSettings,
+      mfaEnabled: true
+    )
+    
+    // Create Facebook provider with Limited Login enabled by default
+    let fbProvider = FacebookProviderSwift()
+    fbProvider.setLimitedLogin(true)
+    facebookProvider = fbProvider
```
russellwheatley: I'm not keen on these changes to authorise user tracking. It should still be self contained within Facebook. 

We should possibly remove the limited login and user tracking all together and check under the hood to see if user tracking is allowed. If so, use classic login, if it isn't allowed, use limited login.

samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExample/App/ContentView.swift
```diff
@@ -0,0 +1,175 @@
...
+                handleLimitedLoginToggle(newValue)
+              }
+            
+            Text("Limited Login protects privacy by preventing Facebook tracking across apps.\nNOTE: THIS IS NOT PART OF THE SDK")
+              .font(.caption)
+              .foregroundColor(.secondary)
+          }
+          .padding()
+        }
+      }
```
russellwheatley: This should all be removed as mentioned above. Should just be `AuthPickerView().environment(authService)`.

samples/swiftui/FirebaseSwiftUIExample/FirebaseSwiftUIExample/App/ContentView.swift
demolaf: fair point

demolaf: I think another thing to note is that _isLimitedLogin_ config in FacebookProviderSwift. if limited login is disabled by the developer and the user denies ATT this would mean facebook sign in is disabled entirely.

PR #1291, opened 2025-10-30
refactor: handle auto upgrade anonymous improvement

russellwheatley:
- Handle both `credentialAlreadyInUse` and `emailAlreadyInUse` when attempting to link anonymous account with provider (if option set in AuthConfiguration).
- Ensure currentUser is `nil` when signedOut for immediate feedback in auth service.
- Created an **opinionated** sign-in handler in AuthPickerView  for **default Views only** which will attempt to link anonymous account with provider and If it fails, it will sign out anonymous user and login with provider. This sign out & login with provider behaviour is not part of AuthService for consumers who don't want it, this means they can get credential from the error thrown (see .accountMergeConflict(context) enum which passes back the credential). 
- Also refactored phone auth to be self contained when starting phone auth flow. Removes the need for random `setVerificationCode()` method.

### DIFF
--- a/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
+++ b/FirebaseSwiftUI/FirebaseAuthSwiftUI/Sources/Views/AuthPickerView.swift
@@ -154,7 +154,10 @@ extension AuthPickerView: View {
             .aspectRatio(contentMode: .fit)
             .frame(width: 100, height: 100)
           if authService.emailSignInEnabled {
-            EmailAuthView().environment(\.signInWithMergeConflictHandler, signInWithMergeConflictHandling)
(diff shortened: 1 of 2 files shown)


FirebaseSwiftUI/FirebaseAppleSwiftUI/Sources/Views/SignInWithAppleButton.swift
```diff
@@ -29,7 +30,13 @@ extension SignInWithAppleButton: View {
   public var body: some View {
     Button(action: {
       Task {
-        try? await authService.signIn(provider)
+        if let handler = signInHandler {
```
demolaf: not necessary but you can do `if let signInHandler {}`

PR #1296, opened 2025-11-06
refactor: allow providers to use whatever sign-in methods they wish

russellwheatley:
## Type Safety Improvements
- Introduced `CredentialAuthProviderSwift` protocol to distinguish credential-based providers from other auth flows
- Provider initializers now accept concrete types (e.g., `GoogleProviderSwift`) instead of generic protocols
- Eliminated force-casting throughout the codebase

## Phone Auth Simplification  
- Removed `PhoneAuthProviderSwift` protocol - phone verification logic now handled internally by `AuthService`
- Simplified phone auth button initialization (no longer requires provider parameter)
- Phone reauthentication now explicitly requires sign-out/sign-in flow

@demolaf - not possible to move phone auth methods out of auth service package because the methods are used in the Views of this package.


demolaf (approved):
LGTM

PR #1302, opened 2025-11-07
fix: allow user to set their own callback for phone auth

russellwheatley:
- Allow callback to withPhoneSignIn() so consumer can push to whatever Phone Auth View they wish.
- Allow callback to withEmailSignIn() so consumer can push to whatever Email Auth View (and/or link) they wish.
- Remove `navigator.push(.mfaResolution)` from AuthService and move to View layer.


demolaf (approved):
LGTM

PR #1328, opened 2026-03-20
docs: getting started for auth firebaseUI using SwiftUI

russellwheatley:
- Wrote a getting started document for the latest FirebaseUI iOS SwiftUI to mirror closely the previous: https://firebase.google.com/docs/auth/ios/firebaseui


morganchen12 (approved):
LGTM with two minor suggestions. Thanks!

dackers86 (approved):

PR #1339, opened 2026-05-04
Configure "delete account" button behavior

megastep:
This makes the "Delete Account" button in the SignedInView configurable, for cases when the application needs more advanced account deletion configuration:

- Existing behavior (default) with confirmation dialog;
- Button disabled; or
- Custom action on clicking

This also includes the changes from my previous PR #1338 to disable the MFA button if that is turned off in the project's configuration.

Also fixes some missing localized strings in the dialog, even though the strings were already localized.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ DerivedData
 .build/
 # Third Party
 /sdk
+vendor/
 
 # Generated
 /target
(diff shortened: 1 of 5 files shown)


russellwheatley:
Hey @megastep - This PR I'm not so sure on. The philosophy behind FirebaseUI Auth SwiftUI - the consumer should either accept the default Views or write their own custom Views, and this is customising the default Views. I'll have to get back to you on this once I've conferred with the team.

russellwheatley:
Thank you for you PR, but after consideration, I'm going to close this out because it doesn't fit in with the philosophy of firebaseUI-iOS. The idea is that the consumer either accepts the default Views or writes their own custom Views.

PR #1349, opened 2026-06-08
Fix stale naming

just1and0:
Update Stale to stale in stale-issue.yml


mikehardy (approved):
just to make sure there are no case-sensitivity issues in expected label name -> ✅

demolaf (approved):

PR #1368, opened 2026-07-07
fix zizmor findings

morganchen12:


morganchen12:
Ignoring failure since this change does not touch test files.

PR #1398, opened 2026-09-18
Preserve AuthService in the authentication sheet environment

### LINKED ISSUE #1397: AuthPickerView sheet loses AuthService environment on Mac Catalyst
### Environment

- Swift, using `FirebaseAuthSwiftUI`, `FirebaseAppleSwiftUI`, and `FirebaseGoogleSwiftUI` through Swift Package Manager
- FirebaseUI 16.1.0
- Firebase 11.12.0; GoogleSignIn 7.1.0
- Mac Catalyst on macOS 26.6.2, Xcode 26.6 (17F113), SDK 26.5, Swift 6.3.3
- Comparison run: iPhone 17 Pro Simulator, iOS 26.5

### Reproduction

In a UIKit-hosted SwiftUI view, use the documented stock picker pattern:

```swift
AuthPickerView {
  Text("Minimal FirebaseUI host")
}
.environment(authService)
```

Configure `authService` with Apple and Google providers, attach the hosting controller to a window, then set `authService.isPresented = true`.

Our hosted XCTest reproduces a fatal error on Catalyst while presenting the sheet:

```
No Observable object of type AuthService found.
A View.environmentObject(_:) for AuthService may be missing as an ancestor of this view.
```

The same stock-picker presentation test passes on iOS Simulator. The Catalyst failure also reproduces with a plain SwiftUI sheet containing an `@Environment(AuthService.self)` reader, suggesting a sheet environment propagation problem rather than provider configuration. Waiting until the presenting content is rendered did not avoid the crash.

### Proposed fix

Explicitly inject `.environment(authService)` into `AuthPickerView`'s primary sheet content, after `.mfaHandler()`. The nested legacy recovery sheet already injects it explicitly.

Both a temporary prototype and the actual patched package passed presentation, dismissal, and simulated successful-authentication handling on iOS Simulator and Catalyst. The consuming app ran six focused hosted tests per platform with zero failures. Real Apple/Google OAuth and the FirebaseUI repository's full test suites were not exercised. I am submitting the corresponding patch.

MatthewCallery:
Fixes #1397.

Presenting `AuthPickerView` from a UIKit-hosted SwiftUI view on Mac Catalyst can crash because the sheet's descendants cannot find `AuthService`, even when the picker has `.environment(authService)`. Explicitly pass the existing service into the primary sheet after its account-conflict and MFA modifiers, matching the explicit injection already used for the nested legacy recovery sheet.

Validation in a consuming UIKit/SwiftUI app, using the patched package based on FirebaseUI 16.1.0:

- Six focused hosted tests passed on iPhone 17 Pro Simulator (iOS 26.5).
- The same six tests passed on Mac Catalyst (macOS 26.6.2).
- The presentation regression checks Apple/Google provider labels, dismissal without account preparation, and simulated successful authentication reaching account preparation.
- The unpatched picker reproduced the missing-`AuthService` crash on Catalyst and passed on iOS.

Built with Xcode 26.6 (17F113). Real provider OAuth and this repository's full test suites were not run. The diff is one line; no dependency versions are changed.


demolaf (approved):
Looks good.

There's a flaky CI run that's meant to be fixed in this PR https://github.com/firebase/FirebaseUI-iOS/pull/1394, I'll try to get that merged and then this PR can be rebased.

demolaf:
@MatthewCallery can you rebase onto `main`? That should fix the CI fail on this PR.

russellwheatley (approved):

